PR #5, opened 2016-09-01
api: add new API flb_service_set()

nokute78:
I modified API document to explain new API `flb_service_set()` which is added https://github.com/fluent/fluent-bit/pull/106

Signed-off-by: Takahiro YAMASHITA nokute78@gmail.com


nokute78:
I fixed conflict. Could you check this patch?

edsiper:
thanks

PR #6, opened 2016-10-11
development: modify return value of API

nokute78:
I modified return values of library API.
They return a negative number like `open(2)`

Signed-off-by: Takahiro YAMASHITA <nokute78@gmail.com>


edsiper:
thanks

PR #13, opened 2017-01-30
in_proc: add 'mem' and 'fd' options and update sample output

nokute78:
I updated a documentation about in_proc.

* update sample output.
* add description about "fd" and "mem" options.


nokute78:
@edsiper ping

edsiper:
thanks

PR #14, opened 2017-03-02
configuration: add a description about filter

nokute78:
I added a explanation about filter plugin to document about configuration file.


edsiper:
thanks

PR #15, opened 2017-03-03
in_mem:update example and add information about swap

nokute78:
I updated a documentation about in_mem.

I added example of this feature.
https://github.com/fluent/fluent-bit/pull/164


edsiper:
thanks

PR #22, opened 2017-04-27
out_file: add description about file format

nokute78:
I added a description about file format which out_file supports.

See also
https://github.com/fluent/fluent-bit/pull/241


edsiper:
thanks

PR #23, opened 2017-05-11
out_file: add description about ltsv format

nokute78:
I added a document about ltsv format.

See also https://github.com/fluent/fluent-bit/pull/255


edsiper:
thanks

PR #30, opened 2017-07-06
filter_parser: add new documentation about filter_parser

nokute78:
I added a documentation about filter_parser.


edsiper:
thanks

PR #40, opened 2017-09-28
Remove instruction to create build directory

JAORMX:
In the build and install instructions, there is no longer the need to
create the "build" directory, since this is already included in the
fluent-bit source.


edsiper:
thanks

PR #48, opened 2017-12-18
Update kubernetes filter docs

wkruse:
A couple of Kubernetes filter config options didn't make it to the docs. The descriptions are mostly from the commit messages. Related to https://github.com/fluent/fluent-bit/issues/367.


wkruse:
TLS settings are already documented http://fluentbit.io/documentation/0.12/configuration/tls_ssl.html, but there is no connection to the Kubernetes filter plugin. Does the Kubernetes filter support all of the TLS settings? Then we could just add a link.

edsiper:
thanks

PR #60, opened 2018-03-27
Fix Fractional Seconds note

martinssipenko:


edsiper:
please make sure to set "section" to be modified in the commit title, e.g:

```
parser: fix fractional seconds note
```

PR #67, opened 2018-05-02
Update README.md

Projoke:
important I think, haha


edsiper:
thanks. fix is already in place.

PR #71, opened 2018-05-11
configuation: file: support wildcard to include files

nokute78:
I added an explanation about wildcard to include multiple files.

https://github.com/fluent/fluent-bit/issues/556
https://github.com/monkey/monkey/pull/282


edsiper:
thanks

PR #82, opened 2018-06-21
Update information regarding Topic_Key

AkashKrDutta:
The `tested` behaviour for the Topic_Key is that: if the value of Topic_Key doesn't matches with any of the values in the Topics list, a default value (i.e. the first topic) from the list is returned. The snippet of code that shows this behaviour is as follows:
```
/* Get first topic of the list (default topic) */
struct flb_kafka_topic *flb_kafka_topic_default(struct flb_kafka *ctx)
struct flb_kafka_topic *flb_kafka_topic_lookup(char *name,int name_len,struct flb_kafka *ctx)
  [...]                      /* No matches, return the default topic */
    return flb_kafka_topic_default(ctx);

}
```
Signed-off-by: Akash Kumar Dutta <akdboltzmann@gmail.com>
Thanks :)


edsiper:
would you please rebase the changes on top of master ?

edsiper:
sorry for the (long) delay, please resolve the conflicts.

AkashKrDutta:
Seems you have already incorporated my changes without merging this branch. :p

PR #93, opened 2018-09-03
add details about splnuk output format

zlangbert:
Adds more information about the output format of the Splunk plugin and adds an example of using `Splunk_Send_Raw` to customize event metadata.


edsiper:
thanks!

PR #133, opened 2019-02-05
Update parser.md on preserving fields

jsravn:
Document option `Preserve_Key`, and improve documentation for `Reserve_Data`.


edsiper:
thanks

PR #134, opened 2019-02-06
Add `Timestamp_Format` in Kafka output

donbowman:
Kafka pipelines into e.g. Druid prefer ISO8601 format over 'Double'

Signed-off-by: Don Bowman <db@donbowman.ca>


edsiper:
thanks

PR #153, opened 2019-03-25
Document Type_int_key for multiple keys

lorenmh:
From looking at the source code, up to 16 keys can be used. Updated to reflect this.


edsiper:
thanks

PR #163, opened 2019-04-17
Update supported resource types in the docs

ensonic:
See relevant code here:
https://github.com/fluent/fluent-bit/blame/master/plugins/out_stackdriver/stackdriver.c#L384


edsiper:
thanks

Issue #151, opened 2019-03-04
[filter_parser] preserve_key and reserve_data documentation incorrect


jamesgoodhouse:
The documentation for how to use the preserve_key and reserve_data options for the parser filter incorrectly show configuring the actual parser with these options and not the filter parser which is where they actually belong.

edsiper:
hmm I don't see the error, would you please point me out what's wrong ? or maybe it got fixed recently ?

https://docs.fluentbit.io/manual/filter/parser

lecaros:
solved in https://github.com/fluent/fluent-bit-docs/pull/173

PR #173, opened 2019-05-27
filter: parser: Reserve_Data should be activated in FILTER

Kwelity:
As noted by #151, the documentation is incorrect.
Reserved_Data and Preserved_Key should be activated in the [FILTER] block instead of the [PARSER] block.

Also, Preserved_Key only works when Reserved_Data is On.
This is why I've swap the block in the documentation.

Relevant part in the code: https://github.com/fluent/fluent-bit/blob/master/plugins/filter_parser/filter_parser.c
```
                            if (ctx->reserve_data) {
                                if (!ctx->preserve_key) {
                                    append_arr_i--;
                                    append_arr_len--;
                                    append_arr[append_arr_i] = NULL;
                                }
                            }
                            else {
                                continue_parsing = FLB_FALSE;
                                break;
                            }
```

I don't actually know what the code do, but I couldn't get the documentation's example working and preserve_key seems to be only invoked here.

### DIFF
--- a/filter/parser.md
+++ b/filter/parser.md
@@ -71,14 +71,72 @@ You can see the record `{"data":"100 0.5 true This is example"}` are parsed.
 
 By default, the parser plugin only keeps the parsed fields in its output.
 
-If you enable `Preserve_Key`, the original key field is preserved:
+If you enable `Reserve_Data`, all other fields are preserved:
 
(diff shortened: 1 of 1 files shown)


edsiper:
thanks

PR #175, opened 2019-06-04
Update k8s workflow example w/ correct value

chiefy:
The current example breaks the filter


edsiper:
thanks. 

closing due to:

- code examples is already fixed in master
- no DCO.

PR #179, opened 2019-06-18
add description about ltsv and logfmt parsers

shogo82148:
ltsv and logfmt parsers are already supported, but there are no document.

- Add new parser: ltsv https://github.com/fluent/fluent-bit/pull/646
- Add new parser: logfmt https://github.com/fluent/fluent-bit/pull/871

I added the document about them.


edsiper:
thanks

PR #187, opened 2019-07-04
Maybe there is a problem in order

arrowfeng:
Signed-off-by: arrowfeng <289716347@qq.com>


fujimotos:
@arrowfeng Thank you for the contribution.

To pass the DCO check, you need to fix the signed-off line using
`git commit --ammend` as follows.

```
before: Signed-off-by: arrowfeng 289716347@qq.com
after : Signed-off-by: HALO Feng 289716347@qq.com
```

PR #193, opened 2019-07-22
installation: add the list of Windows installers (and their checksums)

fujimotos:
* This adds a list of available installers to the documentation.
* Checksums are meant for users who want to check the integrity of
  downloaded packages.


edsiper:
thanks

PR #201, opened 2019-08-13
parser: regular_expression: Add warning about special characters

fujimotos:
This is a limitation of Onigumo, who throws an initialization error
if a group name contains characters other than alphanumeric.

Let's document the limitation in the manual page clearly so that
users do not get confused about it.


fujimotos:
The original issue is https://github.com/fluent/fluent-bit/issues/1483

edsiper:
thanks

PR #202, opened 2019-08-18
input: clarify that both disk and netif are throughput oriented

nigels-com:
(Arguably) Disk usage is usually referring to utilisation of available space (is my disk nearly full), rather than read and write throughput.


edsiper:
thanks

PR #211, opened 2019-09-16
parser: specify parsers which support Types option

nokute78:
To make clear which parser supports `Types` option.

https://github.com/fluent/fluent-bit/issues/1563

> Clearly states in the documentation that this is working as intended and it can be used only with the other parser (like the Decode_Fieldkey that on the contrary work only on json parsers).


edsiper:
thanks

PR #220, opened 2019-10-06
output: http: Sample config and query for Sumo Logic usage

nigels-com:
I did some integration of fluent-bit into a [Sumo Logic](https://www.sumologic.com) dashboard for a Raspberry Pi version 1.

Aligning the `http` output for Sumo Logic is a little bit non-obvious, so I documented some settings that seem to work well for me.  Specifically the `timestamp` and `json_lines` format.

Screenshot from 2019-10-06 22-06-44


edsiper:
thanks

PR #221, opened 2019-10-07
fix(typos): Make instructions for Azure LogInsights a bit more clear

ThorstenHans:
I've updated the Azure docs a bit. I think it's more accurate if the desired Azure service is explicitly referenced instead of just referring to it as *Azure*


edsiper:
thanks

PR #224, opened 2019-10-14
input: collectd: Fix parameter name and typo

talawahtech:
The parameter name for the address to listen on is "Listen", not "Address": https://github.com/fluent/fluent-bit/blob/master/plugins/in_collectd/in_collectd.c#L93

Signed-off-by: Marc Richards <talawahtech@users.noreply.github.com>


edsiper:
thanks

PR #228, opened 2019-10-23
installation: td-agent-bit: add missing Raspbian stretch support

kenhys:
ref: https://docs.fluentbit.io/manual/installation/raspberry_pi

Signed-off-by: Kentaro Hayashi <hayashi@clear-code.com>


edsiper:
thanks

PR #232, opened 2019-10-30
Update exec documentation to include `buf_size`

vlad-stoian:
This was added in this PR: fluent/fluent-bit#1116


edsiper:
thanks

PR #233, opened 2019-10-31
tail.md: document tag expansion

jgehrcke:
I read the documentation for the tail plugin expecting to find information about what `Tag kube.*` actually means. After being confused for quite a while I came across the concept of tail's "tag expansion", so far only documented in the kubernetes filter section. Let's start to document it here, too!


edsiper:
thanks

PR #244, opened 2019-12-10
Ignore_Older is record based

kilya11:
Based on the comment: 
https://github.com/fluent/fluent-bit/issues/716#issuecomment-414460134


edsiper:
thanks

PR #245, opened 2019-12-12
configuration: file: Add better explanation for `Match` option

fujimotos:
kaizen1 pointed out that the explanation about `Match` is unclear
about what it is supposed to match *against*. This seems to be
somewhat confusing to new users.

This sorts out the description, and clarify the usage of `Match`
and its family options.

The original issue is fluent/fluent-bit/issues/1792


edsiper:
thanks

PR #250, opened 2020-01-06
output: Add an undocumented parameter for Elasticsearch

chaodhib:
Document Trace_Error which outputs to console elasticsearch API calls when there is an error


edsiper:
thanks

PR #253, opened 2020-01-13
in_serial - document limitation of bitrate

konradmichael:
The possible values for "bitrate" are limited, but that is currently not documented.


nigels-com:
I would suggest documenting the supported modes, rather than referring to the source code.

edsiper:
as @nigels-com said, not merging since this should be fixed by including the expected values, no referring to the source code. 

we need a new PR to address this. closing this one.

PR #254, opened 2020-01-14
output: forward: Document 'Empty_Shared_Key'

fujimotos:
This option is a special configuration to connect to Fluentd nodes
that use zero-length strings (a.k.a. "") as shared secrets. Let's
document it.


edsiper:
thanks!

PR #262, opened 2020-01-29
Add AWS installation instructions

PettitWesley:


edsiper:
For installation/aws-container.md , would you please instead list in a table the available Docker images and its versions?, on that way the users can get it right away, e.g:

- AWS Fluent Bit image for CloudWatch

PettitWesley:
@edsiper I added more details. In our official docs, we do not actually publish a list of images or versions. Most of our customers use our image through our special AWS repos (not Docker Hub), and for that we build a method by which they can programmatically query for a list of versions and image links for each version. I added instructions for that; these instructions are now in line with the docs in the AWS repo.

installation/docker.md
```diff
@@ -67,5 +67,9 @@ Alpine Linux uses Musl C library instead of Glibc. Musl is not fully compatible
 
 Our Docker containers images are deployed thousands of times per day, we take security and stability very seriously.
 
+<<<<<<< HEAD:installation/docker.md
```
edsiper: looks like a previous git conflict was merged as part of the commit

PettitWesley: Sorry- fixed

edsiper:
thanks

PettitWesley:
@edsiper Since 1.4 has been released, should I open a PR with this for master?

edsiper:
oh, this got into 1.4-dev, sorry about that. it's not being rendered. just did a manual cherry-pick, its ok now. thanks

PR #263, opened 2020-01-29
Add dev guide to docs

PettitWesley:


edsiper:
I think the table of content is not required since gitbook will create one on the left side for every markdown header definition found .

PettitWesley:
@edsiper removed

PR #265, opened 2020-02-12
correct the order of words

earthdiaosi:
correct the order of words


edsiper:
manually fixed and preserved original author. Commit: 3fa620e04

PR #270, opened 2020-02-24
service: remove never merged Config_Watch example

JensErat:
fluent-bit PR [#842] which would have introduced `Config_Watch` was never
merged. This configuration option shouldn't be included in
documentation.

[#842]: https://github.com/fluent/fluent-bit/pull/842


edsiper:
thanks for the contribution. In newer docs the property is not longer listed.

PR #271, opened 2020-02-25
Show the correct default value for `self_hostname`

fujimotos:
While reviewing fluent-bit#1904, I noticed that the current
implementation has a default value for the option ("localhost").
Let's fix it.


edsiper:
thanks, fixed manually on cd7755c (author honored)

PR #272, opened 2020-02-26
configuration: scheduler: Fix the default value of retry_limit to 1

fujimotos:
The documentation states that Fluent Bit performs "2" retries before
giving up by default. The implementation actually uses "1" as the
default value of Retry_Limit. Fix it.


fujimotos:
Originally reported by Zero King at fluent/fluent-bit/issues/1981.

edsiper:
merged manually on d143901

PR #274, opened 2020-03-12
Correct syslog default parser documentation

jon-ruckwood:
I hit this problem today, I thought it was worth expanding the documentation.

---

Selection of the default parser was not as described, refer to:

https://github.com/fluent/fluent-bit/blob/baff15640e97ac46d457aab011e9103f2dca53ce/plugins/in_syslog/syslog_conf.c#L118-L129


edsiper:
thanks. Merged manually on 9075240 (author honored)

PR #276, opened 2020-03-20
Add additional Buffer_Size detail and note

geezyx:
In a recent troubleshooting effort, I had to dig into the source code of fluent-bit to find that `0` was an option for the `Buffer_Size` parameter. I think it's important to document this behavior in this doc.

Additionally, I added a note down in the "Workflow" section which explains the logical flow to give an additional contextual call out of the behavior that folks might run into, which is that much of the pod metadata will not get injected into the record.

I believe this is a silent failure currently, and should probably produce a warning log message since it results in undesirable behavior. I'll add an issue to fluent-bit mentioning this as well.


edsiper:
thanks!

PR #285, opened 2020-04-02
Fix broken documentation link

pgdagenais:
These links are currently broken on the documentation website.

My guess is these are the correct links. 

Thanks


edsiper:
thanks!

PR #287, opened 2020-04-08
installation: sources: build-and-install: Document CMake 3.x

fujimotos:
The default version of CMake on CentOS 7 is 2.x. For this reason,
we keep hearing complaints that cmake fails to set up the build
directory.

Fix that by documenting the version requirement clearly.


fujimotos:
Resolves fluent/fluent-bit/issues/2081

PR #288, opened 2020-04-09
Add instructions to run a logging pipeline locally with docker.

naseemkullah:
It is not uncommon to want to see how a logging pipeline works locally, this can be achieved with docker.

This is, imho, the easiest way for operators to validate logging pipelines locally as to troubleshoot and catch bugs as done in https://github.com/fluent/fluent-bit/issues/2093

### DIFF
--- a/local-testing/logging-pipeline.md
+++ b/local-testing/logging-pipeline.md
@@ -1,0 +1,65 @@
+# Running a Logging Pipeline Locally
+
+You may wish to test a logging pipeline locally to observe how it deals with
+log messages. The following is a walk-through for running Fluent Bit and
+Elasticsearch locally with [Docker Compose](https://docs.docker.com/compose/) which can serve as an example for testing other plugins locally.
+
+## Create a Configuration File
+
+Refer to the [Configuration File
+section](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/configuration-file)
+to create a configuration to test.
+
+`fluent-bit.conf`:
+
+```text
+[INPUT]
+  Name dummy
+  Dummy {"top": {".dotted": "value"}}
+
+[OUTPUT]
+  Name es
+  Host elasticsearch
+  Replace_Dots On
+```
+
+## Docker Compose
+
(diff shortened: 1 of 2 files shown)


development/local-logging-pipeline.md
```diff
@@ -0,0 +1,96 @@
+# Running a logging pipeline locally with Docker
```
PettitWesley, line 1: This needs to be linked somewhere so that it can be found. I'd also remove this sentence "There is no need to setup your local environment for C development
thanks to Docker."

And remove "development" from the file path. 

These are instructions for testing locally using docker.

naseemkullah: Sure.

naseemkullah: Pardon the force push, thanks.

naseemkullah:
Ready for re-review Mr. Wesley

local-testing/logging-pipeline.md
```diff
@@ -0,0 +1,95 @@
...
+
+Use [Docker Compose](https://docs.docker.com/compose/) with the following `docker-compose.yaml`:
+
+```yaml
+version: "3.7"
+
+services:
+  fluent-bit:
+    image: fluent/fluent-bit
+    command:
```
PettitWesley, line 17: I personally don't find the command line arguments for FB to be very convenient... I think it's worthwhile to add a few sentences about how you could use a config file and mount it in on a volume

PettitWesley: Or just change the example to that...

naseemkullah: Good point.

naseemkullah:
Example has been modified to use volume mounts with docker compose, but I've kept cmd line args for the second non-compose example for variety sake.

naseemkullah:
Thanks @PettitWesley, I've doubled down on the compose + config file approach as it is simplest and removed the instructions for ad-hoc docker containers + cmd line args.

@edsiper this is good to merge upon your review/approval.

naseemkullah:
bump

edsiper:
thanks!

PR #290, opened 2020-04-14
pipeline: filter: lua: add description about code 2

nokute78:
I added a description about return code 2.

| return code | description |
|---------------|-------------|
|-1 | drop the record |
|0 | keep the original |
|1| replace timestamp and record |
|2(new)|keep the original timestamp and replace record|

https://github.com/fluent/fluent-bit/pull/2100


edsiper:
thanks

PR #298, opened 2020-04-24
Update azure.md

aSquare14:


aSquare14:
@edsiper This is the correct one I believe

PR #302, opened 2020-05-03
pipeline: filter: lua: add supported version of return code 2

nokute78:
https://github.com/fluent/fluent-bit/pull/2100 is released and supported from v1.4.3.
https://fluentbit.io/announcements/v1.4.3/

I added supported version of the feature.


edsiper:
thanks!

PR #303, opened 2020-05-06
pipeline: filter: lua: add description about protected_mode

nokute78:
filter_lua supports protected mode. https://github.com/fluent/fluent-bit/pull/2154

This is a patch for it.


edsiper:
thanks

PR #304, opened 2020-05-06
pipeline: inputs: dummy: add new property 'Start_time_*sec'

nokute78:
This is a patch for https://github.com/fluent/fluent-bit/pull/2107


nokute78:
@edsiper ping

PR #311, opened 2020-05-22
monitoring: Add the note to enable HTTP server in monitoring docs

championshuttler:
Currently there is no info about it on monitoring docs about how to enable http server while building fluent-bit locally.

cc @edsiper

### DIFF
--- a/administration/monitoring.md
+++ b/administration/monitoring.md
@@ -8,6 +8,11 @@ Fluent Bit comes with a built-in HTTP Server that can be used to query internal 
 
 The monitoring interface can be easily integrated with Prometheus since we support it native format.
 
+## Requirements <a id="requirements"></a>
+
+The monitoring interface depends on the built-in HTTP server feature, make sure this one is available in your build.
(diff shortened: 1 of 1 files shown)


edsiper:
I think the comment must specify that that flag is for Cmake when building Fluent Bit from scratch. Other users using the packages might do a wrong assumption

championshuttler:
> I think the comment must specify that that flag is for Cmake when building Fluent Bit from scratch. Other users using the packages might do a wrong assumption

Done

edsiper:
that's a monitoring section, so the user don't expect to know about cmake. I think is better to refer the user to the build from scratch section

championshuttler:
> that's a monitoring section, so the user don't expect to know about cmake. I think is better to refer the user to the build from scratch section

Should I link `Build from section` to that section?

edsiper:
I would add a header before the getting started like "requirements" mentioning that "The monitoring interface depends on the built-in HTTP server feature, make sure this one is available in your build"

edsiper:
check https://github.com/fluent/fluent-bit-docs/pull/311#issuecomment-655652471

PR #318, opened 2020-06-11
Add note on CRI for kubernetes configmap

egernst:
If a CRI runtime, such as container or CRIO, is utilized, the user should update
the INPUT's parser from docker to cri.

Equivalent PR for kubernetes docs: https://github.com/fluent/fluent-bit-kubernetes-logging/pull/76

Ideally this PR will reference the configmap, rather than the sample flluent-bit repository's parsers.conf

Signed-off-by: Eric Ernst <eric@amperecomputing.com>


edsiper:
thanks

PR #319, opened 2020-06-12
Add Types to regex parser example

qiq:


qiq:
Any update on this? It is helpful to know how to specify types. The information was present in the past, but got lost in a docs refactor.

PR #328, opened 2020-06-26
output:stackdriver: Support special fields - operation

scgao:
Update the doc of Stackdriver - Support special field `operation`.

Signed-off-by: scgao <scgao@google.com>


scgao:
Added description and use cases for `operation`.

PR #333, opened 2020-07-01
pipeline: input: tail: fix Exclude_Path example typo

ich199:
The example for `Exclude_Path` for the `tail` input uses an incorrect syntax.


edsiper:
thanks

PR #342, opened 2020-07-16
Allowed log level is warn and not warning

LionelCons:


dujiulun:
Actually you are referring to the `Log_Level` config item in `[SERVICE]` section, which uses ["warning"](https://github.com/fluent/fluent-bit/blob/960a091acdbf02c53ba112e83fbcedabcf276fb4/src/flb_config.c#L380). For plugins it's indeed ["warn"](https://github.com/fluent/fluent-bit/blob/960a091acdbf02c53ba112e83fbcedabcf276fb4/src/flb_log.c#L180), which is undocumented and a little confusing.

LionelCons:
This is _very_ confusing. Please use the same variant of `warning` everywhere.

edsiper:
thanks

PR #350, opened 2020-07-30
installation: windows: Document "Windows Service"

fujimotos:
This adds a basic explanation on how to register & manipulate
Fluent Bit as Windows Service.

This should be useful for users who want to install Fluent Bit to
their Windows servers.


edsiper:
thanks

PR #351, opened 2020-07-30
out_stackdriver: Document for special fields

scgao:
Add document for special fields of stackdriver output plugin.

Signed-off-by: scgao <scgao@google.com>


edsiper:
thanks

PR #359, opened 2020-08-14
Document multiple arguments to Path

raine:


edsiper:
thanks for your contribution, this changes are already in v1.6 and git master branch. (1.5 is deprecated now).

PR #369, opened 2020-08-26
changed wording for better smoothness / flow

IanWoods1993:
I was reading the docs and found that some of them didn't read as well as they could. The changes are simple - just wording, nothing about the concepts or content itself is wrong. Hope this helps


edsiper:
thanks

PR #375, opened 2020-09-03
out stackdriver: add severity_key and update local_resource_id format

JeffLuoo:
1. add missing option: severity_key
2. update format of local_resource_id
3. add ending dot


JeffLuoo:
1. add missing option: severity_key
2. update format of local_resource_id
3. add ending dot

PR #376, opened 2020-09-10
Fix links in tail and kubernetes doc pages

achals:
Some fixes I found while referring to the docs


achals:
@edsiper mind taking a look at this?

PR #377, opened 2020-09-11
out_es: Document Time_Key_Nanos option

turettn:
Add documentation for new option added in fluent/fluent-bit#2544


edsiper:
thanks

PR #403, opened 2020-10-13
Document FLB_TLS as on by default.

jkschulz:
FLB_TLS is set to "on" [if FLB_OUT_TD is on](https://github.com/fluent/fluent-bit/blob/1b262cebff01a2d54c7dab8ff5d3d8e340faed30/CMakeLists.txt#L247-L249). FLB_OUT_TD is on by default, so FLB_TLS is on by default.


edsiper:
thanks

PR #407, opened 2020-10-22
input: tail: fix default value of Buffer_Max_Size

l2dy:
Even if I set Buffer_Chunk_Size to a higher value, the default Buffer_Max_Size is still 32k.


edsiper:
thanks

PR #429, opened 2020-12-09
Docs: Update cloudwatch.md

SamuelDudley:
Fix example output configuration auto_create_group setting


edsiper:
the current implementation allows both values:

https://github.com/fluent/fluent-bit/blob/master/plugins/out_cloudwatch_logs/cloudwatch_logs.c#L161

no need to change the value on this case

PR #434, opened 2020-12-16
input: winlog: describe 'Tag' option in the documentation

fujimotos:
In fact, WinLog input plugin can handle Tag option without any problem.
For example:

    PS> fluent-bit.exe -i winlog -p tag=test -o stdout

This patch adds the description to make it clear.


fujimotos:
Note: I submit this patch because I saw some users misinterpret it
as "winlog plugin does not support `Tag`".

This patch should expel such confusion.

edsiper:
please resubmit on top of git master

PR #436, opened 2020-12-18
filter/lua: Document record split and many minor improvements

liweitianux:
Add description about that the Lua callback can return an array of
records, i.e., the third return value of the callback function can be an
array of tables.  This feature enables this Lua filter to split an input
record into multiple records.  Document the record split feature as a
new subsection and add an example about it.

See also https://github.com/fluent/fluent-bit/pull/811 .

While there, variosu minor grammar and format fixes.

Signed-off-by: Weitian LI <liweitianux@live.com>

### DIFF
--- a/pipeline/filters/lua.md
+++ b/pipeline/filters/lua.md
@@ -1,11 +1,11 @@
 # Lua
 
-Lua Filter allows you to modify the incoming records using custom [Lua](https://www.lua.org/) Scripts.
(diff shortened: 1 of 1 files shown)


edsiper:
please rebase the specific changes on top of GIT master

liweitianux:
Hi @edsiper , I've rebased my patch against the master.  Please review and handle it.  Thanks.

lecaros:
Thanks for your contribution!

PR #437, opened 2020-12-21
Fix the link to the Unit Sizes

mizukmb:
`Unit Size` link is wrong in forward page.

https://docs.fluentbit.io/manual/pipeline/inputs/forward


agup006 (approved):
Thanks for the PR @mizukmb! Much appreciated

PR #439, opened 2020-12-22
[INPUT] [Tail] Update the warning for DB parameter

vavdoshka:
faced the issue which could be avoided if I saw this warning.


agup006 (approved):

agup006:
Thanks for the PR @vavdoshka, much appreciated!

PR #443, opened 2021-01-13
Fix Service config table to be an actual table

passcod:
Before

image

After

image


agup006 (approved):

agup006:
Awesome thanks for the PR!

PR #450, opened 2021-01-22
Document Elasticsearch Trace_error flag

akosdudas:
The Elasticsearch output plugin has a `Trace_error` flag, which was documented in the past (e.g., https://docs.fluentbit.io/manual/v/1.3/output/elasticsearch), but is not part of the current documentation. The flag still works. Re-added the flag to the docs with its previous text.


agup006:
This shoudl now be fixed, Thank you

PR #452, opened 2021-01-27
out_loki: fix loki output plugin default label

akihiro:
Loki output plugin use default label `job=fluent-bit`

https://github.com/fluent/fluent-bit/blob/26426262ebdf606139af4a241e25ece3afe5451a/plugins/out_loki/loki.c#L373-L381


edsiper:
thanks

PR #453, opened 2021-02-04
add hint of forward plugin not allowing tag assignments to forward input section

xellsys:
The hint that the forward input plugin does not allow Tag assignment is only provided on the general input section but not on the section of the plugin itself. This PR adds this info to the section where a user would expect it.


edsiper:
this PR contains multiple commits, please rebase the specific change on top of GIT master (sorry for the delay)

esmerel:
This is so old it would be safer to redo it if the changes are still needed. There a bunch of files in here we've removed or rewritten.

PR #454, opened 2021-02-04
improve wording of rewrite_tag recrod re-emitting

xellsys:
It is easily missed that the record under the new tag is re-emitted from the beginning of the pipeline. Also I aligned the usage of upper case for the keyword "Rule" to not be easily confused with the "Match rule".


edsiper:
thanks

PR #489, opened 2021-03-11
Allow insecure/unverified S3 endpoint

dtaniwaki:
This is the documentation update for the MinIO support.


mkerins:
Please add `port` to the list of options.

dtaniwaki:
@mkerins I didn't add port because it's not directly configurable by config with the change of my PR. It can be set only by the endpoint config. I didn't add the port config because I thought it would be confusing if you can set it by both the endpoint and port config. What do you think?

mkerins:
> @mkerins I didn't add `port` because it's not directly configurable by config with the change of my PR. It can be set only by the `endpoint` config. I didn't add the `port` config because I thought it would be confusing if you can set it by both the `endpoint` and `port` config. What do you think?

Ah yes, thank you @dtaniwaki that makes sense.

PettitWesley:
@dtaniwaki 

1. I was expecting you to add a section to the docs on how to use MinIO. I think some users would love that.
2. Please make the commit message be descriptive for what you're changing.

dtaniwaki:
@PettitWesley I updated the PR, would you check it?

pipeline/outputs/s3.md
```diff
@@ -119,6 +123,24 @@ Example:
 
 If you enable a single worker, you are enabling a dedicated thread for your S3 output. We recommend starting without workers, evaluating the performance, and then enabling a worker if needed. For most users, the plugin can provide sufficient throughput without workers.
 
+### Usage w/ MinIO
```
PettitWesley, line 126: nit: I like to be more formal in documentation writing, so no "w/"

Is that abbreviation completely universal? Would any user anywhere in the world who understood english enough to read this understand that? 

Another thought is that translation software probably deals better with full words not abbreviations.

dtaniwaki: Fixed. Thank you for telling me that.

PettitWesley (approved):
We can merge this once the feature is released. Ping me when that happens.

edsiper:
note: to be merge once 1.9 is out

lecaros:
Thanks for the contribution! The docs have been merged.

PR #491, opened 2021-03-15
networking: add connection backoff config options

kabakaev:
This PR can be merged only if the [FLB PR 3191](https://github.com/fluent/fluent-bit/pull/3191) is accepted and merged.

### DIFF
--- a/administration/networking.md
+++ b/administration/networking.md
@@ -54,6 +54,10 @@ If TCP keepalive is used, `net.tcp_keepalive_probes` allows to override the OS d
 
 If a TCP connection is keepalive enabled and has very high traffic, the connection may _never_ be killed. In a situation where the remote endpoint is load-balanced in some way, this may lead to an unequal distribution of traffic. Setting `net.keepalive_max_recycle` causes keepalive connections to be recycled after a number of messages are sent over that connection. Once this limit is reached, the connection is terminated gracefully, and a new connection will be created for subsequent messages.
 
+### Connection backoff
+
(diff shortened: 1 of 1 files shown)


edsiper:
note: waiting on https://github.com/fluent/fluent-bit/pull/3191

administration/networking.md
```diff
@@ -54,6 +54,10 @@ If TCP keepalive is used, `net.tcp_keepalive_probes` allows to override the OS d
 
 If a TCP connection is keepalive enabled and has very high traffic, the connection may _never_ be killed. In a situation where the remote endpoint is load-balanced in some way, this may lead to an unequal distribution of traffic. Setting `net.keepalive_max_recycle` causes keepalive connections to be recycled after a number of messages are sent over that connection. Once this limit is reached, the connection is terminated gracefully, and a new connection will be created for subsequent messages.
 
+### Connection backoff
+
+This set of options is designed for large-scale fluent-bit deployments, which may need to limit number of TCP SYN packets in case an output destination is unreachable. By default, fluent-bit will retry connecting to an output target on each chunk flush. But if, for example, `net.initial_backoff=8` and `net.max_backoff=60` options are set, then an output plugin will try to reconnect to a failing output only after a delay of 8, 16, 32 and 60 seconds. Each fluent-bit thread will keep trying only once every 60 seconds, until connection is healthy again.
```
esmerel, line 59:
```suggestion
This set of options is designed for large-scale Fluent Bit deployments, which may need to limit number of TCP SYN packets in case an output destination is unreachable. By default, Fluent Bit will retry connecting to an output target on each chunk flush. But if, for example, `net.initial_backoff=8` and `net.max_backoff=60` options are set, then an output plugin will try to reconnect to a failing output only after a delay of 8, 16, 32 and 60 seconds. Each Fluent Bit thread will keep trying only once every 60 seconds, until connection is healthy again.
```

esmerel: Looks generally ok to me as long as my suggestion is applied, but I can't see the conflicts to fix them.

eschabell:
@kabakaev as you can see, the conflicts I had to fix resulted in taking the path to a new docs PR. I'll update the code PR to reflect that and your attribution will remain in this new docs PR. I will close this in favor of the new #2590 which will be waiting only on the code PR https://github.com/fluent/fluent-bit/pull/3191 merging.

PR #498, opened 2021-03-25
in_dummy: add description about Samples property

nokute78:
I added a description about `Samples` option of `in_dummy`.
It has not been documented  yet.


agup006 (approved):
Looks good from my side

PR #501, opened 2021-04-05
parsers_regex: fix broken links

SanyaKochhar:
Signed-off-by: Sanya Kochhar <kochhars@microsoft.com>

### DIFF
--- a/pipeline/parsers/regular-expression.md
+++ b/pipeline/parsers/regular-expression.md
@@ -6,7 +6,7 @@ Fluent Bit uses [Onigmo](https://github.com/k-takata/Onigmo) regular expression 
 
 [http://rubular.com/](http://rubular.com/)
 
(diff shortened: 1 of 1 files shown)


nokute78:
Your patch is to link a specific commit https://github.com/fluent/fluent-bit-docs/tree/1787fd8bfb2035bf10faf8cb7b14c4521e1265b3.
So, if tail.md will be updated and the commit hash will be changed, the document will remain linking old tail.md.

How about fix like relative link? e.g. `../inputs/tail.md`

SanyaKochhar:
@nokute78 yep, that makes sense, updated to relative links. Wasn't sure if this pointed to a specific commit on purpose before for versioning docs or something

SanyaKochhar:
@edsiper could I get a review on this?

edsiper:
Thanks!

PR #502, opened 2021-04-06
outputs: elasticsearch: add property 'id_key'

nokute78:
I added option `id_key`.

The option is added by https://github.com/fluent/fluent-bit/pull/3320


edsiper:
thanks

PR #533, opened 2021-05-18
windows: remove setx statement

hongaar:
Executing `setx /m path` is dangerous. I've had various team members ruin their shell environment because it erased their Path env var. See e.g. https://stackoverflow.com/a/28778358/938297

This PR changes `setx /m path` to a description to add a directory to the environment variable Path in a non-destructive way.


edsiper:
thanks

PR #545, opened 2021-06-11
filter: kubernetes: add docs for kube_meta_cache_ttl(#3623)

zhonghui12:
Signed-off-by: Zhonghui Hu <zh0512xx@gmail.com>


PettitWesley (approved):

pipeline/filters/kubernetes.md
```diff
@@ -48,6 +48,7 @@ The plugin supports the following configuration parameters:
 | DNS\_Wait\_Time | DNS lookup interval between network status checks | 30 |
 | Use\_Kubelet | this is an optional feature flag to get metadata information from kubelet instead of calling Kube Server API to enhance the log. This could mitigate the [Kube API heavy traffic issue for large cluster](kubernetes.md#optional-feature-using-kubelet-to-get-metadata). | Off |
 | Kubelet\_Port | kubelet port using for HTTP request, this only works when `Use_Kubelet`  set to On. | 10250 |
+| Kube\_Meta\_Cache\_TTL | configurable TTL for K8s cached metadata. | 0 |
```
PettitWesley, line 51: I think the default value should actually be a sentence something like: "TTL for cache entries is disabled by default; cache entries are evicted at random when capacity is reached"

Because the default is not 0. It's "off". We should give a good explanation of the default behavior I think. (My sentence might not be completely accurate, fix it if needed)

PettitWesley (approved):

PR #559, opened 2021-06-26
Correct reference to Azure access keys vs Azure shared keys

artisticcheese:
There is no concept of "Shared Keys" in Azure, there is Shared Access Signatures and there are Access Keys. Current version of Azure blob output is relying on Access Keys (which is essentially `root` keys and referring to them as `shared keys` just confuses users since they would think those are Shared Access Signatures


edsiper:
@artisticcheese thanks for your contribution.

In order to merge it, would you please sign-off your commit with `-s` so DCO error will pass ?

artisticcheese:
I don't know what that means. What sign-off with `-s` means

lecaros:
Hi @artisticcheese, please refer to https://github.com/fluent/fluent-bit-docs/blob/master/CONTRIBUTING.md#sign-off-your-commits for details on signing-off.

PR #561, opened 2021-06-27
out_logdna: added log_key confguration option

svyatogor:


edsiper:
note: waiting for feedback on https://github.com/fluent/fluent-bit/pull/3696

pipeline/outputs/logdna.md
```diff
@@ -78,6 +78,11 @@ Before to get started with the plugin configuration, make sure to obtain the pro
         if not found, the default value is used.</td>
       <td style="text-align:left">Fluent Bit</td>
     </tr>
+    <tr>
+      <td style="text-align:left">log_key</td>
+      <td style="text-align:left">By default, the whole log record will be sent to LogDNA. If you specify a key name with this option, then only the value of  that key will be sent to LogDNA. For example, if you are using  the Fluentd Docker log driver, you can specify log_key log and only  the log message will be sent to LogDNA.</td>
```
esmerel, line 83:
```suggestion
      <td style="text-align:left">By default, the whole log record will be sent to LogDNA. If you specify a key name with this option, then only the value of that key will be sent to LogDNA. For example, if you are using the Fluentd Docker log driver, you can specify `log_key` log and only the log message will be sent to LogDNA.</td>
```

esmerel: "For example, if you are using the Fluentd Docker log driver, you can specify `log_key` log and only the log message will be sent to LogDNA."

This sentence doesn't quite make sense. Is this  "Only the FluentD Docker log's messages" or .. I'm not sure how to parse this.

This PR is also a much older format - it would likely be easier to redo a PR because this one uses HTML and the current format is markdown.

eschabell:
@svyatogor would you want to fix the changes here to the current markdown or start over?

PR #568, opened 2021-07-12
Minor grammer improvements

dbuos:
usage of an and a


agup006:
Thanks for the grammar fixes

PR #570, opened 2021-07-13
pipeline: outputs: syslog: add structured data example

CherylFlowers:
Update the Syslog output doc:
- Add example of how to configure the structured data key.
- Update description.
- Update configuration parameters list to maintain consistency across all config options.


edsiper:
thank you! (sorry for the delay)

PR #573, opened 2021-07-19
aws: add documentation for credentials chain

rittneje:
Added documentation for the AWS credential chain precedence hierarchy, including the recently added support for `credential_process`.

cc @PettitWesley

### DIFF
--- a/administration/aws-credentials.md
+++ b/administration/aws-credentials.md
@@ -1,0 +1,38 @@
+# AWS Credentials
+
+Plugins that interact with AWS services will fetch credentials from various providers in the following order.
+Only the first provider that is able to provide credentials will be used.
+
+All AWS plugins additionally support a `role_arn` (or `AWS_ROLE_ARN`, for [Elasticsearch](../pipeline/outputs/elasticsearch.md)) configuration parameter. If specified, the fetched credentials will then be used to assume the given role.
+
+## 1. Environment Variables
+
+Uses the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (and optionally `AWS_SESSION_TOKEN`) environment variables if set.
+
+## 2. Shared Configuration and Credentials Files
+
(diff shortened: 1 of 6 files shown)


rittneje:
@PettitWesley I noticed that the code has some logic around `EKS_POD_EXECUTION_ROLE`. In particular, if that environment variable is set, then two things happen:
1. the EKS web identity token provider is disabled
2. it uses STS to assume that role using the credentials from the normal chain

https://github.com/fluent/fluent-bit/blob/37aa680d32384c1179f02ee08a5bef4cd278513e/src/aws/flb_aws_credentials.c#L282-L291

However, I don't fully understand why that is happening. I don't use Fargate, but the documentation seems to say that the pod execution role is for the Fargate infrastructure to pull images and such. Containers should still be using web identity tokens as usual. https://docs.aws.amazon.com/eks/latest/userguide/pod-execution-role.html

> The containers running in the Fargate pod cannot assume the IAM permissions associated with the pod execution role. To give the containers in your Fargate pod permissions to access other AWS services, you must use IAM roles for service accounts. 

Is this a bug in Fluent Bit? Should the logic around `EKS_POD_EXECUTION_ROLE` be removed?

PettitWesley:
@rittneje Yeah I believe that is old code left over from the original launch of Fluent Bit support for EKS Fargate. I need to double check that it can be removed but either way it doesn't need to be documented here.

administration/aws-credentials.md
```diff
@@ -0,0 +1,36 @@
+# AWS Credentials
+
+Plugins that interact with AWS services will fetch credentials from various providers in the following order.
+Only the first provider that is able to provide credentials will be used.
```
PettitWesley, line 4: May be add a sentence saying that you can additionally use credentials from an additional IAM Role via the `role_arn`/`aws_role_arn` fields in all AWS plugins. However, what is shown here is the standard credential provider chain and it is used as the base creds to assume that role.

PettitWesley (approved):
@rittneje Thank you for writing this!

administration/aws-credentials.md
rittneje: Added a note for this. Also it seems that GitHub did not revoke your approval when I pushed.

PR #577, opened 2021-07-23
Added documentation for the DNS mode switch and prometheus remote write

leonardo-albertovich:


edsiper:
@leonardo-albertovich there are some recent conflicts, please resolve them so we can merge it

PR #582, opened 2021-07-28
Update fluentd-and-fluent-bit.md

nbon12:
The wording "But now is more than a simple tool" was confusing to me.
Is it saying that fluentd has become more than a simple tool? or is it saying that fluentbit has become more than a simple tool?


agup006:
Thanks for submitting @nbon12 ! This change helps explain that sentence much better. Would you be able to update DCO so we can merge

PR #586, opened 2021-08-11
Update to file output plugin configuration

kishaningithub:


agup006:
Hi @kishaningithub thanks for the pull request, though I don't see the benefit it is adding for a user

PR #601, opened 2021-09-09
Adds troubleshooting for Roles for AKS clusters

headconnect:
Adding some details on potentially azure-specific rolebinding issues where Azure doesn't permit kubernetes filter to retrieve data due to not being a member of `system:serviceaccounts`.

Regardless, also adding a quick way to check if the permissions are correctly set with the `can-i` command :)

### DIFF
--- a/pipeline/filters/kubernetes.md
+++ b/pipeline/filters/kubernetes.md
@@ -381,6 +381,26 @@ If you are not seeing metadata added to your kubernetes logs and see the followi
 
 When Fluent Bit is deployed as a DaemonSet it generally runs with specific roles that allow the application to talk to the Kubernetes API server. If you are deployed in a more restricted environment check that all the Kubernetes roles are set correctly.
 
+You can test this by running the following command (replace `fluentbit-system` with the namespace where your fluentbit is installed)
+```text
+kubectl auth can-i list pods --as=system:serviceaccount:fluentbit-system:fluentbit
+```
+If set roles are configured correctly, it should simply respond with `yes`. 
+
+For instance, using Azure AKS, running the above command may respond with: 
+```text
+no - Azure does not have opinion for this user.
+```
+
+If you have connectivity to the API server, but still "could not get meta for POD" - debug logging might give you a message with `Azure does not have opinion for this user`. Then the following `subject` may need to be included in the `fluentbit` `ClusterRoleBinding`: 
+
+appended to `subjects` array:
+```yaml
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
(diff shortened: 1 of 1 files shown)


pipeline/filters/kubernetes.md
```diff
@@ -381,6 +381,26 @@ If you are not seeing metadata added to your kubernetes logs and see the followi
...
 
+You can test this by running the following command (replace `fluentbit-system` with the namespace where your fluentbit is installed)
+```text
+kubectl auth can-i list pods --as=system:serviceaccount:fluentbit-system:fluentbit
+```
+If set roles are configured correctly, it should simply respond with `yes`. 
+
+For instance, using Azure AKS, running the above command may respond with: 
+```text
+no - Azure does not have opinion for this user.
```
patrick-stephens, line 392: Is this the real message? It seems to be missing the word "an" before "opinion" so just want to make sure we get the exact (current) message.

pipeline/filters/kubernetes.md
```diff
@@ -381,6 +381,26 @@ If you are not seeing metadata added to your kubernetes logs and see the followi
...
+```
+
+If you have connectivity to the API server, but still "could not get meta for POD" - debug logging might give you a message with `Azure does not have opinion for this user`. Then the following `subject` may need to be included in the `fluentbit` `ClusterRoleBinding`:
```
patrick-stephens, line 395: Can we add any links to AKS docs (if they have them)?

pipeline/filters/kubernetes.md
headconnect: That is actually the real text - copypasted verbatim from my terminal. Don't hate the player, hate the.. cloud provider? ;)

pipeline/filters/kubernetes.md
headconnect: I figured this out through this issue: https://github.com/Azure/AKS/issues/954
That issue also asks for a reference to documentation, though none was provided..

pipeline/filters/kubernetes.md
patrick-stephens: Always! 😄

pipeline/filters/kubernetes.md
patrick-stephens: Should we always include that subject? Or does it work sometimes?

edsiper:
@patrick-stephens is this ready to go ?

patrick-stephens (approved):

patrick-stephens:
Yes, I can't merge though @edsiper

PR #611, opened 2021-09-24
Delete duplicated record

com30n:
the doc has a duplicated options tag_prefix


agup006:
Thanks for the cleanup!

PR #612, opened 2021-09-27
possible typo in `listen` vs. `host`

shantanugadgil:
possible typo (listen vs. host)


shantanugadgil:
@agup006 tagging for visibility

edsiper:
`host` is used to representing a host we are connecting to, instead of `listen` for when an interface needs to bind an address (server) mode, so this change is not appropiated. 

if there is a functionality bug, pls elaborate. closing for now

PR #621, opened 2021-09-30
Add all avaliable vars for compilation

tguenneguez:
Add all avaliable vars for compilation


edsiper:
Invalid PR, it just removes an empty line

PR #626, opened 2021-10-14
Add documentation for the NGINX metrics input plugin.

pwhelan:


pwhelan:
Add documentation from nginx exporter with reference and attribution to the original project.

PR #636, opened 2021-11-17
output: elasticsearch: Document how to fix validation errors

fujimotos:
This adds a FAQ that explains how to fix the following error:

    {"type":"action_request_validation_exception",
     "reason":"Validation Failed: 1: an id must be
     provided if version type or value are set;"}

Should fix fluent/fluent-bit/issues/3909


edsiper:
thanks

PR #640, opened 2021-11-29
Alert NoOutputBytesProcessed should combine outputs

MPV:
We were seeing issues where we got alerted due to some outputs not sending data for more than 15 minutes. 

For example, this is expected from journald logs on worker nodes in a cluster, thus this adjustment to the alert.

We now aggregate output per pod and alert if no bytes are sent for 15 minutes.


edsiper:
- I would suggest to extend alerts for kubernetes only, note the current example is just for vanilla Fluent Bit
- you have to sign-off the commits on this PR

edsiper:
- I would suggest to extend alerts for kubernetes only, note the current example is just for vanilla Fluent Bit
- you have to sign-off the commits on this PR

lecaros:
Hi, @MPV would you mind reviewing the last comment?
Thanks!

MPV:
@lecaros could you review this again? commit has been signed now.

MPV:
@edsiper? 😊

PR #641, opened 2021-12-01
Build flag `FLB_OUT_NATS` is default to On

nonylene:
Build flag `FLB_OUT_NATS` seems to be enabled by default, according to the current CMakeLists.txt:
https://github.com/fluent/fluent-bit/blob/8c491202d62d996b7412726bb2c13f0761b887f6/CMakeLists.txt#L159


edsiper:
thanks

PR #642, opened 2021-12-01
Fix Authorization header in remote-write example

annettejanewilson:
The example for sending prometheus metrics to New Relic sets the header "bearer_token" when it should be setting the header "Authorization". The former doesn't work, and results in a 400 response. The Logz.io and Coralogix examples already used the correct form, this just brings the main example into line.


edsiper:
thanks!

PR #644, opened 2021-12-02
Add documentation for fluentbit_metrics plugin

annettejanewilson:
Hi, I wanted to export fluentbit's own metrics via Prometheus remote-write. I spent quite a while looking at the [Fluent Bit monitoring docs][1] and scratching my head until a colleague pointed out that the `fluentbit_metrics` input plugin provides these metrics.

The `fluentbit_metrics` input plugin was undocumented. It's valuable for emitting these metrics via remote-write. This PR adds documentation similar to the documentation on the `node_exporter` input plugin.

I've manually tested the configuration in the example. I note that the node_exporter plugin [example][2] which I based it on fails due to having "listen" instead of "host" as a configuration key name.

Is this plugin considered public, are these docs useful? They would have helped me.

Are they okay as they are? I worry the example might be a bit confusing since just configuring a prometheus_exporter output doesn't achieve anything that couldn't already be done with the configuration explained under [monitoring][1] but a remote-write setup would not make a nice standalone example.

[1]: https://docs.fluentbit.io/manual/administration/monitoring
[2]: https://docs.fluentbit.io/manual/pipeline/inputs/node-exporter-metrics#simple-configuration-file

### DIFF
--- a/pipeline/inputs/fluentbit-metrics.md
+++ b/pipeline/inputs/fluentbit-metrics.md
@@ -1,0 +1,60 @@
+---
+description: >-
+  A plugin to collect Fluent Bit's own metrics
+---
+
+# Fluent Bit Metrics
+
+Fluent Bit exposes its [own metrics](../../administration/monitoring.md) to allow you to monitor the internals of your pipeline.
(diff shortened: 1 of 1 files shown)


agup006:
@annettejanewilson thank you so much for this documentation! This is great, merging

PR #693, opened 2022-01-31
docker: update 1.8.12 base image details

patrick-stephens:
Addresses https://github.com/fluent/fluent-bit/issues/4645#issuecomment-1025464523 by ensuring we have the Debian 11 update in there.

Signed-off-by: Patrick Stephens <pat@calyptia.com>


lecaros (approved):

jerome-leanspace:
I just realized it still says Busybox on the dockerhub (https://hub.docker.com/r/fluent/fluent-bit/), maybe this can also be addressed?

patrick-stephens:
@lecaros that was changed, there's no reason for busybox as the debug images are full shell ones already plus it broke the exec plugin by installing busybox so it was removed from 1.8.12+

I updated the docs but not the docker hub page for it

patrick-stephens:
@jerome-leanspace just trying to figure out where the text for DockerHub comes from and I'll update once I do...

PR #714, opened 2022-02-17
Filter nightfall docs

victor88121:


lecaros:
Thanks for the contribution @victor88121 !

PR #725, opened 2022-02-25
clarify health check with more details

DrewZhang13:
clarify health check with more details.
Include tow part:

1. A equation for how to determine health status
2. a example with the equation
3. note that error count and retry count only based on output plugins and worked for all output plugins

### DIFF
--- a/administration/monitoring.md
+++ b/administration/monitoring.md
@@ -220,14 +220,22 @@ Sample alerts are available [here](https://github.com/fluent/fluent-bit-docs/tre
 
 Fluent bit now supports four new configs to set up the health check.
 
-| Config Name            | Description                                                               | Default Value |
-| ---------------------- | ------------------------------------------------------------------------- | ------------- |
-| Health_Check           | enable Health check feature                                               | Off           |
-| HC_Errors_Count        | the error count to meet the unhealthy requirement                         | 5             |
-| HC_Retry_Failure_Count | the retry failure count to meet the unhealthy requirement                 | 5             |
-| HC_Period              | The time period by second to count the error and retry failure data point | 60            |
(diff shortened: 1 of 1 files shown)


PettitWesley (commented):
Thanks, please also edit the option description in the table (above your change), so that its clear in their descriptions that output metrics are being used.

PettitWesley (commented):
Copying a discussion we had here for tracking purposes:


Wesley Pettit  3 minutes ago
So the problem is that I realized not all folks know what an official error/retry is

Wesley Pettit  3 minutes ago
One customer thought that errors meant any error in the fluent bit logs

Wesley Pettit  3 minutes ago
So we should give examples

Wesley Pettit  3 minutes ago
Here's a retry message:
[2022/02/16 20:11:36] [ warn] [engine] chunk '1-1645042288.260516436.flb' cannot be retried: task_id=0, input=tcp.3 > output=cloudwatch_logs.1

Wesley Pettit  2 minutes ago
let me find an example of a real output failure message...

Wesley Pettit  1 minute ago
[2022/02/16 10:44:10] [ warn] [engine] failed to flush chunk '1-1645008245.491540684.flb', retry in 7 seconds: task_id=0, input=forward.1 > output=cloudwatch_logs.3 (out_id=3)

Wesley Pettit  < 1 minute ago
Please add these as examples and explain that this is what its based on

Wesley Pettit  < 1 minute ago
Not any error

Wesley Pettit  < 1 minute ago
These are also warn messages unfortunately which is confusing...

Wesley Pettit  < 1 minute ago
May be we should ask eduardo to change them to error level

PettitWesley (approved):
Thanks!

PR #740, opened 2022-03-13
README: fix stream-processing link

nokute78:
https://docs.fluentbit.io/manual/
```
Stream Processing: Perform data selection and transformation using simple SQL queries
```
Above link links github.com. 
This patch is to modify the link which links to https://docs.fluentbit.io/.


lecaros:
@nokute78 thanks for your contribution! I've merged it.

PR #763, opened 2022-03-24
Document configuring plugin log_level

annettejanewilson:
`Log_Level` can be set for individual plugins the same way it can be set for the service as a whole. Is this supported? If so, could we add this to the docs like this?


patrick-stephens:
@annettejanewilson does this actually work then, I'm not sure if that's what you're saying or requesting?

annettejanewilson:
It does work - at least it did to the best of my knowledge at the time I raised this PR. I did not and do not know why it wasn't documented. Assuming that the lack of documentations was an oversight, I raised this ticket.

PR #783, opened 2022-04-09
filter_checklist: add properties.

nokute78:
This PR is to add below properties.
https://github.com/fluent/fluent-bit/blob/3cd0061de651db61e70573105bfff21df258355c/plugins/filter_checklist/checklist.c#L541-L556

- mode
- print_query_time
- ignore_case


lecaros:
Thanks @nokute78 I've merged it!

PR #784, opened 2022-04-10
in_prom_scrape: modify scrape_interval

nokute78:
I fixed `scrape interval`.
I think it needs underscore. `scrape_interval`.

https://github.com/fluent/fluent-bit/blob/v1.9.2/plugins/in_prometheus_scrape/prom_scrape.c#L194


lecaros:
Thanks @nokute78! I've merged it.

PR #790, opened 2022-04-15
Successful response code

nokute78:
https://docs.fluentbit.io/manual/pipeline/inputs/http
https://docs.fluentbit.io/manual/pipeline/outputs/http

According to the document, out_http supports `successful_response_code` and in_http doesn't.

However the description is opposite.
in_http supports `successful_response_code` and out_http doesn't.
https://github.com/fluent/fluent-bit/blob/v1.9.2/plugins/in_http/http.c#L151


lecaros:
Thanks for catching this @nokute78. I've merged it

PR #800, opened 2022-04-28
out_azure: add parameters time_key and time_generated do docs

strima:


lecaros:
Thanks for your contribution @strima!

PR #814, opened 2022-05-24
input_winevtlog: documentation updates

mdiez-modus:
Two updates, one is the default setting for `String_Inserts` so it matches with that is in the code, the other is a subtle information about `Use_ANSI` that took me a week and reading the source code to find out.

Signed-off-by: Marcos Diez <marcos@unitron.com.br>


lecaros:
Thanks for your contribution @mdiez-modus! I've merged it.

PR #832, opened 2022-06-26
Docs about config files included using wildcard are unsorted

lht:
This patch adds document about the current behavior. The ordering issue is raised in https://github.com/fluent/fluent-bit/issues/5612

Signed-off-by: Haitao Li <lihaitao@gmail.com>

### DIFF
--- a/administration/configuring-fluent-bit/classic-mode/configuration-file.md
+++ b/administration/configuring-fluent-bit/classic-mode/configuration-file.md
@@ -160,3 +160,5 @@ Wildcard character (\*) is supported to include multiple files, e.g:
 ```
 @INCLUDE input_*.conf
 ```
+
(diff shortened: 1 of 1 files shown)


administration/configuring-fluent-bit/classic-mode/configuration-file.md
```diff
@@ -160,3 +160,5 @@ Wildcard character (\*) is supported to include multiple files, e.g:
 ```
 @INCLUDE input_*.conf
 ```
+
+Note files matching the wildcard character are included unsorted. If plugins ordering between files need to be preserved, the files should be incldued explicitly.
```
patrick-stephens, line 164: Split into two lines please.
Also type for `incldued`

patrick-stephens:
```suggestion
Note files matching the wildcard character are included unsorted. 
If plugins ordering between files need to be preserved, the files should be included explicitly.
```

PR #835, opened 2022-07-01
Add docs for multipart upload

andy-wang-f:
Add docs for multipart upload


pipeline/outputs/s3.md
```diff
@@ -18,31 +18,31 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| upload_chunk_size             | The size of each 'part' for multipart uploads without gzip. Max: 50M; for multipart uploads with gzip. Max: 50M                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | 5,242,880 bytes                         |
```
PettitWesley, line 28: language nit/suggestion: "The size of each 'part' for multipart uploads.  For uncompressed uploads, the max is 50M. For compressed uploads, the max is 5G. This is the size of the buffered data before compression is applied, so the actual uploaded chunk will be smaller. The min upload_chunk_size in all cases is restricted in the S3 API to 5,242,880 bytes. If after compression, the chunk is smaller than that size, the plugin will upload the chunk using the PutObject API instead."

Something like this, we need more explanation for the user. Another idea is that we should instead not make this part of the doc too long and complicated and have a special section further down to discuss multipart compression. And then in this part just note that the user should read that section. Let's discuss next week.

pipeline/outputs/s3.md
```diff
@@ -18,31 +18,31 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| compression                   | Compression type for S3 objects. 'gzip' is currently the only supported value. The Content-Encoding HTTP Header will be set to 'gzip'. Compression all type can be enabled when `use_put_object` is on, Compression Gzip can be enabled when `use_put_object` on or off. If Apache Arrow support was enabled at compile time, you can set 'arrow' to this option.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | None                                    |
```
PettitWesley, line 39: let's change this to something like: 

"Compression type for S3 objects. 'gzip' is currently the only supported value by default. The Content-Encoding HTTP Header will be set to 'gzip'.  If Apache Arrow support was enabled at compile time, you can set 'arrow' to this option. Compression 'gzip' can be enabled with both `use_put_object` on or off. Apache 'arrow' compression can only be enabled with use_put_object` on." 

I think this is more readable.

andy-wang-f: Will do

pipeline/outputs/s3.md
andy-wang-f: Will do

PettitWesley (approved):

PettitWesley:
Closing in favor of: https://github.com/fluent/fluent-bit-docs/pull/991

PR #848, opened 2022-07-19
pipeline: parsers: configuring-parser: add docs for skip_empty_values

### LINKED ISSUE #847: parsers: skip_empty_values is not documented
I've noticed that `skip_empty_values` is not mentioned in the documentation as a possible configuration parameter. There are situations in which a multiline rule expects an empty line, such as in golang stack traces and this option makes it possible to use a regex parser that will correctly handle such a case for structured inputs.

Looking at the [implementation](https://sourcegraph.com/github.com/fluent/fluent-bit/-/blob/src/flb_parser.c?L515:9), it seems like supporting this option through the public API was intentional. If this is the case, I think that it would be beneficial to add it to the public docs as well.

YoussefRizkalla:
Closes #847

Signed-off-by: YoussefRizkalla <youssefr@google.com>


YoussefRizkalla:
CC: @JeffLuoo @courageJ

patrick-stephens (approved):

PR #857, opened 2022-08-02
outputs: loki output add worker configuration paramenter

toledompm:
`Worker` parameter is missing from most outputs documentation, copied it from [stdout docs](https://docs.fluentbit.io/manual/pipeline/outputs/standard-output). I recently discovered it worked by digging through some github issues. Is this config available for all outputs? And if so is there a more centralized doc where this should be / already is available?


lecaros:
hi @toledompm,
you can use the `workers` parameter in Loki, however, it doesn't have a default value of 1.
This is available for all output plugins.

toledompm:
>it doesn't have a default value of 1.

I skimmed over some [source code](https://github.com/fluent/fluent-bit/blob/1.7/src/flb_output.c#L953), is the default 0?

> This is available for all output plugins.

Should there be a general output doc? Recently we ran into some issues with backpressure that got fixed after we increased our output workers, but we only found out that was a thing after seeing it mentioned in the [Kinesis output docs](https://docs.fluentbit.io/manual/pipeline/outputs/kinesis#worker-support)

lecaros:
hi @toledompm the default is no workers. Meaning it will run on the same thread the other components are running.
We currently don't have documentation for Outputs in general; that's a good idea. We always welcome PRs :)

PR #859, opened 2022-08-08
Fix Helm Chart reference

wandersonwhcr:
This PR fixes a Helm Chart reference.

It is a simple typo.


patrick-stephens:
Thanks for that @wandersonwhcr

PR #868, opened 2022-08-22
buffering-and-storage: add a few notes about how pause_on_chunks_overlimit works.

pwhelan:


lecaros:
Thanks @pwhelan! Merged.

PR #896, opened 2022-09-20
Bump dependency inside local-pipeline

petemounce:
The previous version listed does not appear to be in the registry any longer. This is the latest release in the 7.x series of Elasticsearch.


petemounce:
@lecaros thanks for the link; I've read it now. I see it's the DCO. Since I made the commit via a driveby in the github UI, what's the best way to rectify?

PR #899, opened 2022-09-21
Add another use case to docs

francescsanjuanmrf:
Add another use case for lua scripts.

Istio's proxies logs response code filtering use case.


francescsanjuanmrf:
Will reopen in another PR

PR #903, opened 2022-09-21
pipeline:filters:multiline-stacktrace removed incorrect entry for ruby support from list

RicardoAAD:
…y support from list

Signed-off-by: RicardoAAD <ricardo@calyptia.com>


lecaros:
Thanks for your contribution @RicardoAAD! I've merged it

PR #912, opened 2022-09-29
Typo correction

elenacbc:


lecaros:
Thanks for these contributions @elenacbc! I've merged them

PR #936, opened 2022-10-15
Specifying nginx_plus off for nginx metrics

pablokbs:
By default is enabled so we need to disable if we are not using Nginx Plus

Signed-off-by: Pablo Fredrikson <pablo.fredrikson@gmail.com>


lecaros:
Hi @pablokbs 
thanks for this PR.
Could you please review our contributing guidelines and modify the message accordingly?
https://github.com/fluent/fluent-bit-docs/blob/master/CONTRIBUTING.md#commit-subjects

We'll gladly merge the PR after that.

PR #941, opened 2022-10-19
pipeline: output: kafka: Update docs

deanvz:
Updating the windows docs with how to on kafka output plugin for windows builds as the openssl reference is wrong


pipeline/outputs/kafka.md
```diff
@@ -47,6 +47,45 @@ In your main configuration file append the following _Input_ & _Output_ sections
...
+
+```text
+LINK : warning LNK4044: unrecognized option '/lpthread'; ignored
+LINK : fatal error LNK1104: cannot open file 'libcrypto.lib'
+NMAKE : fatal error U1077: '"C:\Program Files\CMake\bin\cmake.exe"' : return code '0xffffffff'
+```
+
+This is due to the fact that the location of the openssl library required is incorrect. You can fix it by following the steps below
+
+Update the file <fluentbit_repo>/cmake/windows-setup.cmake
```
patrick-stephens, line 64: You can actually do it via the `-DFLB_WINDOWS_DEFAULTS=No` to disable the automatic set up of Windows defaults and then manually turn it on/off as you have below. This is probably the better way to do it rather than having to edit source code.

https://github.com/fluent/fluent-bit/blob/b913c38bc2bf1f61792a177d5820cfc2689a4203/CMakeLists.txt#L131

pipeline/outputs/kafka.md
```diff
@@ -47,6 +47,45 @@ In your main configuration file append the following _Input_ & _Output_ sections
...
+
+The kafka output plugin will produce an error similar to the following when trying to build for windows
+
+```text
+LINK : warning LNK4044: unrecognized option '/lpthread'; ignored
+LINK : fatal error LNK1104: cannot open file 'libcrypto.lib'
+NMAKE : fatal error U1077: '"C:\Program Files\CMake\bin\cmake.exe"' : return code '0xffffffff'
+```
+
+This is due to the fact that the location of the openssl library required is incorrect. You can fix it by following the steps below
```
patrick-stephens, line 62: I think you've missed the actual steps to install/modify OpenSSL installation directory.

patrick-stephens (changes requested):
Thanks for this, I think you've missed the actual OpenSSL change though.

Can you split sentences to a separate line? It makes it easier to comment on.

PR #956, opened 2022-10-25
pipeline:out_s3: added store_dir_limit_size  documentation

Claych:
Added store_dir_limit_size to s3 configuration parameters

### DIFF
--- a/installation/docker.md
+++ b/installation/docker.md
@@ -16,6 +16,8 @@ The following table describes the tags that are available on Docker Hub [fluent/
 
 | Tag(s)      | Manifest Architectures    | Description                                                  |
 | ----------- | ------------------------- | ------------------------------------------------------------ |
+| 2.0.5       | x86\_64, arm64v8, arm32v7 | Release [v2.0.5](https://fluentbit.io/announcements/v2.0.5/) |
+| 2.0.5-debug | x86\_64, arm64v8, arm32v7 | v2.0.x releases (production + debug)                         |
 | 2.0.4       | x86\_64, arm64v8, arm32v7 | Release [v2.0.4](https://fluentbit.io/announcements/v2.0.4/) |
 | 2.0.4-debug | x86\_64, arm64v8, arm32v7 | v2.0.x releases (production + debug)                         |
 | 2.0.3       | x86\_64, arm64v8, arm32v7 | Release [v2.0.3](https://fluentbit.io/announcements/v2.0.3/) |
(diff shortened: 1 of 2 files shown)


pipeline/outputs/s3.md
```diff
@@ -28,6 +28,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| store_dir_limit_size                        | The size of the limitation for disk usage in S3. Limit the amount of s3 buffers in the `store_dir` to limit disk usage.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                | 0, which means unlimited                              |
```
PettitWesley, line 31: let's add a note that this config must be used instead of the normal `storage.total_limit_size` parameter that they are used to for other plugins

pipeline/outputs/s3.md
```diff
@@ -28,6 +28,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| store_dir_limit_size                        | The size of the limitation for disk usage in S3. Limit the amount of s3 buffers in the `store_dir` to limit disk usage. This configuration must be used instead of the normal `storage.total_limit_size` parameter that they are used to for other plugins.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | 0, which means unlimited                              |
```
PettitWesley, line 31: > This configuration must be used instead of the normal `storage.total_limit_size` parameter that they are used to for other plugins.  

Do you think this language will be very clear to a customer and is the most ideal? Who is they? Can we reword this?

pipeline/outputs/s3.md
```diff
@@ -28,6 +28,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| store_dir_limit_size                        | The size of the limitation for disk usage in S3. Limit the amount of s3 buffers in the `store_dir` to limit disk usage. Note: if you want to upload data to S3, using `store_dir_limit_size` instead of `storage.total_limit_size` which is used for other plugins.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | 0, which means unlimited                              |
```
PettitWesley, line 31: Sorry I think we can say this more succinctly and also explain it with:

```
S3 has its own buffering system, so use `store_dir_limit_size` instead of `storage.total_limit_size`.
```

PettitWesley (approved):

PR #977, opened 2022-11-18
administration: scheduling-and-retries: add retry waiting time section

DrewZhang13:
Signed-off-by: Drew Zhang <zhayuzhu@amazon.com>

### DIFF
--- a/administration/scheduling-and-retries.md
+++ b/administration/scheduling-and-retries.md
@@ -10,6 +10,65 @@ Once an output plugin gets called to flush some data, after processing that data
 
 If the return status was **OK**, it means it was successfully able to process and flush the data, if it returned an **Error** status, means that an unrecoverable error happened and the engine should not try to flush that data again. If a **Retry** was requested, the _Engine_ will ask the _Scheduler_ to retry to flush that data, the Scheduler will decide how many seconds to wait before that happens.
 
+## Configuring Wait Time for Retry
+
+The Scheduler provides two configuration options called **scheduler.cap** and **scheduler.base** which can be set in the Service section.
+
+| Key | Description | Default Value | 
+| -- | ------------| --------------| 
+| scheduler.cap | Set a maximum retry time in seconds. The property is supported from v1.8.7. | 2000 | 
+| scheduler.base | Set a base of exponential backoff. The property is supported from v1.8.7. | 5 |
+
+These two configuration options determine the waiting time before a retry will happen. 
+
+Fluent Bit uses an exponential backoff and jitter algorithm to determine the waiting time before a retry.
+
+The waiting time is a random number between a configurable upper and lower bound.
+
+For the Nth retry, the lower bound of the random number will be:
+
+`base`
+
+The upper bound will be:
+
+`min(base * (Nth power of 2), cap)`
+
+Given an example where `base` is set to 3 and `cap` is set to 30. 
+
+1st retry: The lower bound will be 3, the upper bound will be 3 * 2 = 6. So the waiting time will be a random number between (3, 6).
+
+2nd retry: the lower bound will be 3, the upper bound will be 3 * (2 * 2) = 12. So the waiting time will be a random number between (3, 12).
+
+3rd retry: the lower bound will be 3, the upper bound will be 3 * (2 * 2 * 2) = 24. So the waiting time will be a random number between (3, 24).
+
+4th retry: the lower bound will be 3, since 3 * (2 * 2 * 2 * 2) = 48 > 30, the upper bound will be 30. So the waiting time will be a random number between (3, 30).
+
+Basically, the **scheduler.base** determines the lower bound of time between each retry and the **scheduler.cap** determines the upper bound.
+
+For a detailed explanation of the exponential backoff and jitter algorithm, please check this [blog](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).
+
+### Example
+
+The following example configures the **scheduler.base** as 3 seconds and **scheduler.cap** as 30 seconds. 
+
+```text
+[SERVICE]
+    Flush            5
+    Daemon           off
+    Log_Level        debug
+    scheduler.base   3
+    scheduler.cap    30
+```
+
+The waiting time will be:
+
+| Nth retry | waiting time range (seconds) |
+| --- | --- | 
+| 1 | (3, 6)  |
+| 2 | (3, 12) |
+| 3 | (3, 24) |
+| 4 | (3, 30) |
+
 ## Configuring Retries
 
 The Scheduler provides a simple configuration option called **Retry\_Limit** which can be set independently on each output section. This option allows to disable retries or impose a limit to try N times and then discard the data after reaching that limit:
--- a/administration/configuring-fluent-bit/classic-mode/configuration-file.md
+++ b/administration/configuring-fluent-bit/classic-mode/configuration-file.md
@@ -48,6 +48,7 @@ The following is an example of a _SERVICE_ section:
     Daemon          off
     Log_Level       debug
 ```
+For scheduler and retry details, please check there: [scheduling and retries](../../scheduling-and-retries.md#Scheduling-and-Retries)
 
 ## Input <a href="config_input" id="config_input"></a>
 
--- a/administration/configuring-fluent-bit/yaml/configuration-file.md
+++ b/administration/configuring-fluent-bit/yaml/configuration-file.md
@@ -65,6 +65,8 @@ service:
     log_level: debug
 ```
 
+For scheduler and retry details, please check there: [scheduling and retries](../../scheduling-and-retries.md#Scheduling-and-Retries)
+
 ## Pipeline <a href="config_pipeline" id="config_pipeline"></a>
 
 A _pipeline_ section will define a complete pipeline configuration, including _inputs_, _filters_ and _outputs_ subsections.


administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,41 @@ Once an output plugin gets called to flush some data, after processing that data
...
+```
+
+The waiting time will be:
+
+| Nth retry | waiting time range (seconds) |
+| --- | --- | 
+| 1 | (5, 16)   |
+| 2 | (5*2, 16) |
+| 3 | (16, 16)  |
+| 4 | (16, 16)  |
```
matthewfala, line 46: Hmm, are you sure about this? From the blog post you submitted it seems like jitter at least in AWS' context means we will always select a random number between 0 and the current exponentually backed off wait time. So it would be 

0-5
0-10
0-16
0-16

Now I'm not sure if that is what the fluent bit code is doing but I believe that is what the amazon article expects it to do. Please let me know if you have a different understanding.

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,41 @@ Once an output plugin gets called to flush some data, after processing that data
...
 If the return status was **OK**, it means it was successfully able to process and flush the data, if it returned an **Error** status, means that an unrecoverable error happened and the engine should not try to flush that data again. If a **Retry** was requested, the _Engine_ will ask the _Scheduler_ to retry to flush that data, the Scheduler will decide how many seconds to wait before that happens.
 
+## Configuring Wait Time for Retry
+
+The Scheduler provides two configuration options called **scheduler.cap** and **scheduler.base** which can be set on service section. 
+
+| Key | Description | Default Value | 
+| -- | ------------| --------------| 
+| scheduler.cap | Set a maximum retry time in second. The property is supported from v1.8.7. | 2000 | 
+| scheduler.base | Set a base of exponential backoff. The property is supported from v1.8.7. | 5 |
```
PettitWesley, line 20: These are also mentioned in other places: 
- https://github.com/fluent/fluent-bit-docs/blob/master/administration/configuring-fluent-bit/classic-mode/configuration-file.md
- https://github.com/fluent/fluent-bit-docs/blob/master/administration/configuring-fluent-bit/yaml/configuration-file.md

@DrewZhang13 I think we need to make sure that any place the user finds these settings, they are directed to this description.  So I think we need a link or update there too.

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,41 @@ Once an output plugin gets called to flush some data, after processing that data
...
+
+These two configuration determine the waiting time before retry happen. Fluent Bit use exponential-backoff-and-jitter algorithm to determine the waiting time. These configuration means when the Nth retry needed, scheduler will generate a random number X between `(min(base * (Nth power of 2), cap), cap)` as waiting time for this retry.Basically, the **scheduler.base** decide how fast retry could happen, the **scheduler.cap** decide what's the longest waiting time retry could be.
```
matthewfala, line 22: Revise wording?


>These two configurations determine the waiting time before retry attempts. Fluent Bit uses exponential-backoff-and-jitter algorithm to determine the waiting time. These configurations mean that when the Nth retry is needed, scheduler will generate a random number X between `(min(base * (Nth power of 2), cap), cap)` as the waiting time for this retry. Basically, **scheduler.base** decides how fast retries can happen, while **scheduler.cap** decides what's the longest time each retry can be delayed.

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,41 @@ Once an output plugin gets called to flush some data, after processing that data
...
+| scheduler.cap | Set a maximum retry time in second. The property is supported from v1.8.7. | 2000 | 
+| scheduler.base | Set a base of exponential backoff. The property is supported from v1.8.7. | 5 |
+
+These two configuration determine the waiting time before retry happen. Fluent Bit use exponential-backoff-and-jitter algorithm to determine the waiting time. These configuration means when the Nth retry needed, scheduler will generate a random number X between `(min(base * (Nth power of 2), cap), cap)` as waiting time for this retry.Basically, the **scheduler.base** decide how fast retry could happen, the **scheduler.cap** decide what's the longest waiting time retry could be. 
+
+For detail explanation on exponential-backoff-and-jitter algorithm, please check this [blog](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).
+
+### Example
+
+The following example configure the **scheduler.base** as 5 seconds and **scheduler.cap** as 16 seconds.
```
matthewfala, line 28: The following example configure**s**

Add s to configure

matthewfala (changes requested):

administration/scheduling-and-retries.md
PettitWesley, line 22: Can you give an example calculation to help folks understand? Also I think the formula on its own line is best.

Also I actually do not know what this formula means:

> scheduler will generate a random number X between `(min(base * (Nth power of 2), cap), cap)`

What are the upper and lower bounds for the random number generation? Is the `cap` the upper bound? I think we need a more clear way to write this.

PettitWesley: The example below does help

PettitWesley: Separate the upper and lower bound same as you do in the example, it makes it clearer

administration/scheduling-and-retries.md
DrewZhang13: Yeah you are right, it should be the upper bound keep changing until reach to the maximum, the lower bound will be the base. Which the range should be:
5 - 10
5 - 16
5 - 16
5 - 16

administration/scheduling-and-retries.md
DrewZhang13: Good point, will add links in those places for reference

administration/scheduling-and-retries.md
DrewZhang13: Ack, thanks!

DrewZhang13: Will separate the formula into independent line and add an example following it. 
Basically, lower bound would be the base, and upper bound will keep increasing with retry time goes until reach to the cap.

PettitWesley (approved):
Thanks! @patrick-stephens FYI

administration/scheduling-and-retries.md
matthewfala: The logic looks good now.

matthewfala (approved):

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,61 @@ Once an output plugin gets called to flush some data, after processing that data
...
+
+1st time retry waiting time: The lower bound will be 3, the upper bound will be 3 * 2 = 6. So the waiting time will be a random number between (3, 6).
+
+2nd time retry waiting time: the lower bound will be 3, the upper bound will be 3 * (2 * 2) = 12. So the waiting time will be a random number between (3, 12).
+
+3rd time retry waiting time: the lower bound will be 3, the upper bound will be 3 * (2 * 2 * 2) = 24. So the waiting time will be a random number between (3, 24).
+
+4th time retry waiting time: the lower bound will be 3, since 3 * (2 * 2 * 2 * 2) = 48 > 30, the upper bound will be 30. So the waiting time will be a random number between (3, 30).
+
+Basically, the **scheduler.base** decide the lower bound time retry need to wait, the **scheduler.cap** decide the maximum upper bound time retry need to wait.
```
patrick-stephens, line 42:
```suggestion
Basically, the **scheduler.base** determines the lower bound of time between each retry and the **scheduler.cap** determines the upper bound.
```

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,61 @@ Once an output plugin gets called to flush some data, after processing that data
...
+
+`base`
+
+The upper bound will be:
+
+`min(base * (Nth power of 2), cap)`
+
+Given a example when base is setup to 3, cap is set up to 30. 
+
+1st time retry waiting time: The lower bound will be 3, the upper bound will be 3 * 2 = 6. So the waiting time will be a random number between (3, 6).
```
patrick-stephens, line 34: I'd be tempted to remove the extra text in these bits so just "1st retry:..., 2nd retry:..."

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,61 @@ Once an output plugin gets called to flush some data, after processing that data
...
+
+2nd time retry waiting time: the lower bound will be 3, the upper bound will be 3 * (2 * 2) = 12. So the waiting time will be a random number between (3, 12).
+
+3rd time retry waiting time: the lower bound will be 3, the upper bound will be 3 * (2 * 2 * 2) = 24. So the waiting time will be a random number between (3, 24).
+
+4th time retry waiting time: the lower bound will be 3, since 3 * (2 * 2 * 2 * 2) = 48 > 30, the upper bound will be 30. So the waiting time will be a random number between (3, 30).
+
+Basically, the **scheduler.base** decide the lower bound time retry need to wait, the **scheduler.cap** decide the maximum upper bound time retry need to wait. 
+
+For detail explanation on exponential-backoff-and-jitter algorithm, please check this [blog](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).
```
patrick-stephens, line 44:
```suggestion
For a detailed explanation of the exponential backoff and jitter algorithm, please check this [blog](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/).
```

Can we rely on this blog always being there? Or should we incorporate it?

administration/scheduling-and-retries.md
```diff
@@ -10,6 +10,61 @@ Once an output plugin gets called to flush some data, after processing that data
...
+
+### Example
+
+The following example configure the **scheduler.base** as 3 seconds and **scheduler.cap** as 30 seconds.
```
patrick-stephens, line 48:
```suggestion
The following example configures the **scheduler.base** as 3 seconds and **scheduler.cap** as 30 seconds.
```

patrick-stephens (commented):
Some minor typos I think we should update but otherwise good I think.

DrewZhang13:
@patrick-stephens Thanks for the comments! Addressed them and updated. Please help to take another look, Thanks!

lecaros (approved):

PR #978, opened 2022-11-20
installation:linux-packages:ubuntu add hint about ca-certificates

e-gloo:
Add hint for the following error "Certificate verification failed".


patrick-stephens:
@e-gloo any chance you could add this to the other Linux targets and also the top-level one? Rather than just have it for Ubuntu.

PR #983, opened 2022-11-22
multiline-parsing: add an example to get structured data from multiline

nokute78:
This is a patch for https://github.com/fluent/fluent-bit/pull/6257

PR #994, opened 2022-12-02
multiline-parsing: update regex

nokute78:
This patch is to update regex.
It was requested at https://github.com/fluent/fluent-bit-docs/pull/983#discussion_r1029929372


matschaffer:
Thanks for this!

PR #998, opened 2022-12-06
Remove info hint opening in backpressure

orirawlings:


patrick-stephens:
Also fixes the incorrect linkage

PR #1029, opened 2023-01-19
buffering: fix inaccuracies in buffering and storage doc

PettitWesley:
1. Mem_Buf_Limit only applies when using default memory buffer
2. storage.max_chunks_up is the setting when using filesystem
3. storage.backlog.mem_limit is for chunks recovered from a previous run
4. Misc improvements for better clarity

Hopefully later I will get time to fully rewrite this doc to make it clearer for novice users and also more readable for users who will just skim sections and need to understand how to solve their use cases without reading the full doc.

Signed-off-by: Wesley Pettit <wppttt@amazon.com>


PettitWesley:
@edsiper @pwhelan Can we get this reviewed and merged ASAP? I think these doc inaccuracies are very dangerous.

Claych (approved):

PettitWesley:
Normally I would not merge PRs to a central component that I/my team doesn't own. But in this case, these inaccuracies are dangerous and need to be fixed ASAP. If you provide review I will fix it in a subsequent PR.

PR #1031, opened 2023-01-22
pipeline: filters: grep: add logical_op property

nokute78:
This patch is for https://github.com/fluent/fluent-bit/pull/6721

### DIFF
--- a/pipeline/filters/grep.md
+++ b/pipeline/filters/grep.md
@@ -14,6 +14,7 @@ The plugin supports the following configuration parameters:
 | :--- | :--- | :--- |
 | Regex | KEY  REGEX | Keep records in which the content of KEY matches the regular expression. |
 | Exclude | KEY REGEX | Exclude records in which the content of KEY matches the regular expression. |
+| Logical_Op| Operation | Specify whether to use logical conjuciton or disjunction. `AND` and `OR` are allowed as an Operation.|
 
 #### Record Accessor Enabled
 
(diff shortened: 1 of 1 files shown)


pipeline/filters/grep.md
```diff
@@ -14,6 +14,7 @@ The plugin supports the following configuration parameters:
 | :--- | :--- | :--- |
 | Regex | KEY  REGEX | Keep records in which the content of KEY matches the regular expression. |
 | Exclude | KEY REGEX | Exclude records in which the content of KEY matches the regular expression. |
+| Logical_Op| Operation | Specify whether to use logical conjuciton or disjunction. `AND` , `OR` and `legacy` are allowed as an Operation. `legacy` is for backward compatibility. Default is `legacy`.|
```
lecaros, line 17: What's the behavior when the value is `legacy`?

lecaros (changes requested):

pipeline/filters/grep.md
nokute78: It is for current behavior to keep backward compatibility.
Current behavior is a buggy but user may expect such behavior.
Could you check from https://github.com/fluent/fluent-bit/pull/6721#issuecomment-1407594534 ?

lecaros: I think we should explain what the legacy mode does, even if it's buggy.

nokute78: My opinion is here. It is hard to explain for me.
https://github.com/fluent/fluent-bit/pull/6721#issuecomment-1409492327

Cc: @patrick-stephens

patrick-stephens:
```suggestion
| Logical_Op| Operation | Specify which logical operator to use. `AND` , `OR` and `legacy` are allowed as an Operation. Default is `legacy` for backward compatibility. In `legacy` mode the behaviour is either AND or OR depending whether the `grep` is including (then it's AND) or excluding (then it's OR).|
```

nattawitc:
Hello, there's a minor issue that are very small I'm not sure if I should create a new issue for this, but the example output

```
Fluent Bit v2.0.9
* Copyright (C) 2015-2022 The Fluent Bit Authors
* Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
* https://fluentbit.io
[2023/01/22 09:46:49] [ info] [fluent bit] version=2.0.9, 
```

Shouldn't version be at least 2.1.0? It confuse me a bit when I read the doc since I'm on the same version but it's not working.

patrick-stephens:
@nattawitc that means you are running 2.0.9, not 2.1.0

Make sure you have pulled the latest version.

nattawitc:
@patrick-stephens Yes, I understand that and already pull the latest one. What I mean is that doc is a bit misleading that it's available on v2.0.9. It could be improved by changing the version in an example or some note that it's a new feature on v2.1.0. I understand that it's very minor thing, that's why I commented it here instead of raising an issue.

patrick-stephens:
It does say: `Only available from 2.1+.` in the docs I can see for 2.1 (the default - make sure to select 2.0 docs explicitly).
Screenshot from 2023-04-24 10-17-59

If you go to the 2.0 specific docs then it does not include it: https://docs.fluentbit.io/manual/v/2.0/pipeline/filters/grep
Screenshot from 2023-04-24 10-17-38

PR #1034, opened 2023-01-26
in_node_exporter_metrics: Add descriptions for individual metrics intervals

cosmo0920:
Corresponding PR is https://github.com/fluent/fluent-bit/pull/6737.

Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>


pipeline/inputs/node-exporter-metrics.md
```diff
@@ -22,6 +22,21 @@ This plugin is currently only supported on Linux based operating systems\
...
+| ne.stat.scrape\_interval      | The rate at which stat metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
+| ne.time.scrape\_interval      | The rate at which time metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
+| ne.loadavg.scrape\_interval   | The rate at which loadavg metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
+| ne.vmstat.scrape\_interval   | The rate at which vmstat metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
+| ne.filefd.scrape\_interval   | The rate at which filefd metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
+| metrics | To specify which metrics are collected from the host operating system | `"cpu,cpufreq,meminfo,diskstats,uname,stat,time,loadavg,vmstat,netdev,filefd"` |
```
patrick-stephens, line 35: I believe regardless of which metrics you enable/disable you still have to mount or provide access to the same underlying filesystem components - can we add a note along these lines?

https://github.com/fluent/fluent-bit/pull/6737#issuecomment-1406211983

People may rightly assume that if I turn everything off (or whatever they don't need but as an extreme example) then they can just run an unprivileged container with no extra mounts.

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -22,6 +22,21 @@ This plugin is currently only supported on Linux based operating systems\
...
+
+**Note:** For backward compatibilities of this plugin configurations, `scrape_interval` should behave as a root scraper for node's metrics. This could be affected for scraping intervals.
```
patrick-stephens, line 37: I think what you're saying is that `scrape_interval` is the global default and you can specifically override the value for certain metrics?

```suggestion
**Note:** The plugin top-level `scrape_interval` setting is the global default with any custom settings for individual `scrape_intervals` then overriding just that specific metric scraping interval.
```

Also can we use separate lines for each sentence as easier to review then.

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -22,6 +22,21 @@ This plugin is currently only supported on Linux based operating systems\
...
+Each of `ne.xxx.scrape_interval` can specify each of intervals of metrics.
+However, they cannot affect the root collector of metrics to send the latter plugin pipeline.
```
patrick-stephens, line 39: I'm not sure what this means? It needs rewording I think.

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -22,6 +22,21 @@ This plugin is currently only supported on Linux based operating systems\
 | scrape_interval | The rate at which metrics are collected from the host operating system | 5 seconds |
 | path.procfs     | The mount point used to collect process information and metrics        | /proc/    |
 | path.sysfs      | The path in the filesystem used to collect system metrics              | /sys/     |
+| ne.cpu.scrape\_interval       | The rate at which cpu metrics are collected from the host operating system. To enable, specify greater than 0. | 0 seconds (disabled by default) |
```
patrick-stephens, line 25: Instead of setting to 0 would it be better to have it unset? Or does it require a value?

`Disabled by default` is not right here, I think we should say it uses the global `scrape_interval` by default (what if that is set to 0??). We do not disable the polling of metrics, we just do not configure it with an override.

patrick-stephens (changes requested):

pipeline/inputs/node-exporter-metrics.md
cosmo0920: > People may rightly assume that if I turn everything off (or whatever they don't need but as an extreme example) then they can just run an unprivileged container with no extra mounts.

Sure. BTW, everything turned off should be failed to start. Everything turned off is worthless...

patrick-stephens: Yup but it is possible - particularly if people do dynamic config via substitution or something...
That's the only way to turn on/off outputs for example dynamically using the same config - you do something like:
```
match ${OUTPUT_MATCH}
```
Then ensure the variable is set to whatever you want to match or some un-matcheable string.
I used this approach for the Couchbase sidecar for example as it is a lot simpler to just set an annotation that appears via the DownwardAPI as a variable and have the same super config everywhere - particularly for customers who do not want to learn FB. You have an image with fixed config then, just dynamically setting variables at pod deployment time.

Anyway, the point is someone could disable things but then it fails to run as we still need the things mounted even though they're not used. Let's document that at least clearly.

pipeline/inputs/node-exporter-metrics.md
cosmo0920: `ne.xxx.scrape_interval` will ignite to start each of timers for tied up metrics. They should be disabled -- this is because we should keep as-is or keep plugin behavior at least from behavior of first implementation of this plugin.

On Fluentd world, we can announce through deprecated warning that can be generated from config_param options. However, Fluent Bit does not have such functionality.

pipeline/inputs/node-exporter-metrics.md
cosmo0920: Yes. Sorry for complicated and unclear sentence...

pipeline/inputs/node-exporter-metrics.md
patrick-stephens: Sure, I understand the implementation but the documentation is not clear from a user perspective I think. In that case the global scrape_interval is used: if you do not set a custom interval here then you will get metrics at the global interval right? The user only cares about the outcome and not what it means for the implementation - saying it is disabled suggests they won't get these metrics.

cosmo0920: Ah, you are right. There was lack of description for "specific intervals".
Modified in https://github.com/fluent/fluent-bit-docs/pull/1034/commits/f322923a50346594f36c0ed664473b482fa90b24.

pipeline/inputs/node-exporter-metrics.md
cosmo0920: Ah, dynamically configuring way should be opened up my patches. I didn't realize such use cases like an k8s deployment with environment variables.  Hmm..., I'll try to write down to clarify procfs and sysfs requirements.

pipeline/inputs/node-exporter-metrics.md
cosmo0920: I tried to reword this sentence. https://github.com/fluent/fluent-bit-docs/pull/1034/commits/6fed928b9c2e0c790610a6ec98514862e9f789f3

pipeline/inputs/node-exporter-metrics.md
cosmo0920: I didn't work about this comment. This will be "my homework" for the next week. :smile:

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -15,13 +15,28 @@ The initial release of Node Exporter Metrics contains a subset of collectors and
...
+| collector.loadavg.scrape\_interval   | The rate at which loadavg metrics are collected from the host operating system. To enable for a specific callaction interval, specify greater than 0. | 0 seconds |
+| collector.vmstat.scrape\_interval   | The rate at which vmstat metrics are collected from the host operating system. To enable for a specific collection interval, specify greater than 0. | 0 seconds |
+| collector.filefd.scrape\_interval   | The rate at which filefd metrics are collected from the host operating system. To enable for a specific collection interval, specify greater than 0. | 0 seconds |
+| metrics | To specify which metrics are collected from the host operating system | `"cpu,cpufreq,meminfo,diskstats,filesystem,uname,stat,time,loadavg,vmstat,netdev,filefd"` |
+
+**Note:** The plugin top-level `scrape_interval` setting is the global default with any custom settings for individual `scrape_intervals` then overriding just that specific metric scraping interval.
+Each of `collector.xxx.scrape_interval` can specify each of gathering intervals of metrics.
```
patrick-stephens, line 38:
```suggestion
Each `collector.xxx.scrape_interval` option only overrides the interval for that specific collector and the associated set of metrics provided.
```

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -15,13 +15,28 @@ The initial release of Node Exporter Metrics contains a subset of collectors and
...
+However, all of node metrics will be collected the interval which is specified with the global `scrape_interval`.
```
patrick-stephens, line 39: I would just remove this to be honest, I think you've already said it fine.
```suggestion
```

pipeline/inputs/node-exporter-metrics.md
```diff
@@ -15,13 +15,28 @@ The initial release of Node Exporter Metrics contains a subset of collectors and
...
 
-## Configuration 
+## Configuration
 
 | Key             | Description                                                            | Default   |
 | --------------- | ---------------------------------------------------------------------- | --------- |
 | scrape_interval | The rate at which metrics are collected from the host operating system | 5 seconds |
 | path.procfs     | The mount point used to collect process information and metrics        | /proc/    |
 | path.sysfs      | The path in the filesystem used to collect system metrics              | /sys/     |
+| collector.cpu.scrape\_interval       | The rate at which cpu metrics are collected from the host operating system. To enable for a specific collection interval, specify greater than 0. | 0 seconds |
```
patrick-stephens, line 25: Something like this for each of them?
```suggestion
| collector.cpu.scrape\_interval | The rate in seconds at which cpu metrics are collected from the host operating system. If a value greater than 0 is used then it overrides the global default otherwise the global default is used. | 0 seconds |
```

pipeline/inputs/node-exporter-metrics.md
cosmo0920: I added procfs/sysfs requirements on https://github.com/fluent/fluent-bit-docs/pull/1034/commits/bfd3f7f0e43e9e203fe643aa157cc7376ed7a9f3.

patrick-stephens (approved):

PR #1038, opened 2023-02-02
tail: add note about perf for ignore_older

PettitWesley:
Signed-off-by: Wesley Pettit <wppttt@amazon.com>


Claych (approved):

pipeline/inputs/tail.md
```diff
@@ -19,7 +19,7 @@ The plugin supports the following configuration parameters:
...
-| Ignore\_Older | Ignores files which modification date is older than this time in seconds. Supports m,h,d \(minutes, hours, days\) syntax.                                                                                                                                                                                                                                  |  |
+| Ignore\_Older | Ignores files which modification date is older than this time in seconds. Supports m,h,d \(minutes, hours, days\) syntax. Configuring this option can improve performance when your Path matches a lot of older files that are no longer being actively modified.                                                                                                                                                                                                                                |  |
```
matthewfala, line 22: Are we sure that this resolves the issue yet? Customer compliant about fluent bit is 1 failure per 2 weeks.
Maybe we should wait for confirmation before publishing these doc updates?

PettitWesley: Well this doc update doesn't mention any issue, it just says it improves performance, which is true right? Monitoring fewer files means you have less work to do. 

We are not yet certain of the importance of this config for one specific customer. I think the doc update is not specific to that.

lizsnyder (approved):

pipeline/inputs/tail.md
PettitWesley: Now that we know that really you need both log rotation and this, and log rotation is more important, I will close this PR

PR #1044, opened 2023-02-10
pipeline: outputs: s3: add the description of time in s3_key

Claych:
Signed-off-by: Clay Cheng <claychen@amazon.com>


pipeline/outputs/s3.md
```diff
@@ -31,7 +31,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
...at string for keys in S3. This option supports a UUID, strftime time formatters, a syntax for selecting parts of the Fluent log tag using a syntax inspired by the rewrite\_tag filter. Add $UUID in the format string to insert a random string. Add $INDEX in the format string to insert an integer that increments each upload. The $INDEX value will be saved in the store_dir so that if Fluent Bit restarts the value will keep incrementing from the previous run. Add $TAG in the format string to insert the full log tag; add $TAG\[0] to insert the first part of the tag in the s3 key. The tag is split into “parts” using the characters specified with the `s3_key_format_tag_delimiters` option. Add extension directly after the last piece of the format string to insert a key suffix. If you want to specify a key suffix and you are in `use_put_object` mode, you must specify $UUID as well. More explanations can be found in the S3 Key Format explainer section further down in this document. See the in depth examples and tutorial in the documentation. Time in s3_key is the time for the first log in the file which is recorded as the creation time of the file.  | /fluent-bit-logs/$TAG/%Y/%m/%d/%H/%M/%S |
```
PettitWesley, line 34: > Time in s3_key is the time for the first log in the file which is recorded as the creation time of the file. 

Sorry I think this is confusing and possibly misleading. **It's the timestamp of the first record in the S3 file.** It does not have anything to do with the creation time of the file. The user could be ingesting very old logs which might have timestamps that are days or weeks old, in which case, that's how the S3 key will be set. The current language doesn't make this clear enough.

PettitWesley, line 34: > Time in s3_key is the time for the first log in the file which is recorded as the creation time of the file. 

Sorry I think this is confusing and possibly misleading. **It's the timestamp of the first record in the S3 file.** It does not have anything to do with the creation time of the file. The user could be ingesting very old logs which might have timestamps that are days or weeks old, in which case, that's how the S3 key will be set. The current language doesn't make this clear enough.

PettitWesley (changes requested):

PettitWesley (approved):

PR #1075, opened 2023-04-13
input: stdin: Describe EoF behaviour, parser

ringerc:
Document how to use a parser with the stdin plugin to read arbitrary messages, since the default behaviour is to assume a specific json input format. This is very confusing for new users, since the plugin just logs

    invalid record found, it's not a JSON map or array

when it sees inputs it doesn't expect, and the docs don't explain that you can set a parser to handle other formats.

Document the stdin plugin's behaviour on end-of-file.

Also explain the accepted input formats for the plugin's default reader: object, object with timestamp, or object with timestamp and metadata.

### DIFF
--- a/concepts/key-concepts.md
+++ b/concepts/key-concepts.md
@@ -30,11 +30,35 @@ Jan 18 12:52:16 flb gsd-media-keys[2640]: # watch_fast: "/org/gnome/terminal/leg
 
 It contains four lines and all of them represents **four** independent Events.
 
-Internally, an Event always has two components \(in an array form\):
+Internally an Event is comprised of:
+
+* timestamp
+* key/value metadata (since v2.1.0)
+* payload
+
+### Event format
+
+The Fluent Bit wire protocol represents an Event is a 2-element array
+with a nested array as the first element:
+
+```javascript
+[[TIMESTAMP, METADATA], MESSAGE]
+```
+
+where
+
+* TIMESTAMP is a timestamp in seconds as an integer or floating point value (not a string);
+* METADATA is a possibly-empty object containing event metadata; and
+* MESSAGE is an object containing the event body.
+
+Fluent-bit versions prior to v2.1.0 instead used:
 
 ```javascript
 [TIMESTAMP, MESSAGE]
 ```
+
+to represent events. This format is still supported for reading input event
+streams.
 
 ## Filtering
 
--- a/concepts/key-concepts.md
+++ b/concepts/key-concepts.md
@@ -38,7 +38,7 @@ Internally an Event is comprised of:
 
 ### Event format
 
-The Fluent Bit wire protocol represents an Event is a 2-element array
+The Fluent Bit wire protocol represents an Event as a 2-element array
 with a nested array as the first element:
 
 ```javascript
--- a/concepts/key-concepts.md
+++ b/concepts/key-concepts.md
@@ -51,7 +51,7 @@ where
 * METADATA is a possibly-empty object containing event metadata; and
 * MESSAGE is an object containing the event body.
 
-Fluent-bit versions prior to v2.1.0 instead used:
+Fluent Bit versions prior to v2.1.0 instead used:
 
 ```javascript
 [TIMESTAMP, MESSAGE]
--- a/pipeline/inputs/standard-input.md
+++ b/pipeline/inputs/standard-input.md
@@ -1,17 +1,35 @@
 # Standard Input
 
-The **stdin** plugin allows to retrieve valid JSON text messages over the standard input interface \(stdin\). In order to use it, specify the plugin name as the input, e.g:
+The **stdin** plugin supports retrieving a message stream from the standard input interface \(stdin\) of the Fluent Bit process.
+In order to use it, specify the plugin name as the input, e.g:
 
 ```bash
 $ fluent-bit -i stdin -o stdout
 ```
 
-As input data the _stdin_ plugin recognize the following JSON data formats:
+If the stdin stream is closed (end-of-file), the stdin plugin will instruct
+Fluent Bit to exit with success (0) after flushing any pending output.
+
+## Input formats
+
+If no parser is configured for the stdin plugin, it expects *valid json* input input data with one of the following data formats:
 
 ```bash
 1. { map => val, map => val, map => val }
 2. [ time, { map => val, map => val, map => val } ]
 ```
+
+Any input data that is *not* in one of the above formats will cause the plugin to log errors like:
+
+```
+[error] [input:stdin:stdin.0] invalid record found, it's not a JSON map or array
+```
+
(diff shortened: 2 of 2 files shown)


ringerc:
How do I progress this?

patrick-stephens:
What do you need? I can review.

ringerc:
@patrick-stephens Review would be greatly appreciated, yes. Thankyou.

pipeline/inputs/standard-input.md
```diff
@@ -1,18 +1,37 @@
 # Standard Input
 
-The **stdin** plugin allows to retrieve valid JSON text messages over the standard input interface \(stdin\). In order to use it, specify the plugin name as the input, e.g:
+The **stdin** plugin allows to retrieve a message stream from the standard input interface \(stdin\) of the fluent-bit process. In order to use it, specify the plugin name as the input, e.g:
```
patrick-stephens, line 3: Let's ensure we use `Fluent Bit` everywhere and split sentences to simplify reviews.

```suggestion
The **stdin** plugin supports retrieving a message stream from the standard input interface \(stdin\) of the Fluent Bit process.
In order to use it, specify the plugin name as the input, e.g:
```

pipeline/inputs/standard-input.md
```diff
@@ -1,18 +1,37 @@
...
 $ fluent-bit -i stdin -o stdout
 ```
 
-As input data the _stdin_ plugin recognize the following JSON data formats:
+If the stdin stream is closed (end-of-file), the stdin plugin will instruct
+`fluent-bit` to exit with success (0) after flushing any pending output.
+
+## Input formats
+
+If no parser is configured for the stdin plugin, it expects *valid json* input input data with one of the following data formats:
```
patrick-stephens, line 14:
```suggestion
If no parser is configured for the stdin plugin, it expects *valid json* input data with one of the following data formats:
```

I'm not sure those are valid JSON though - the `=>` is something `stdout` uses?

pipeline/inputs/standard-input.md
```diff
@@ -1,18 +1,37 @@
...
 
 ```bash
 1. { map => val, map => val, map => val }
 2. [ time, { map => val, map => val, map => val } ]
 ```
 
+It will log errors like
```
patrick-stephens, line 21: I'm a little confused here - above it says to expect data in this format then it says there will be an error.

Can we clarify with correct format and then examples of failing ones?

pipeline/inputs/standard-input.md
```diff
@@ -48,11 +67,56 @@ Fluent Bit v1.x.x
...
+$ seq 1 5 | /opt/fluent-bit/bin/fluent-bit -c fluent-bit.conf -R parser.conf -q
+[0] stdin: [1681358780.517029169, {"message"=>"1"}]
+[1] stdin: [1681358780.517068334, {"message"=>"2"}]
+[2] stdin: [1681358780.517072116, {"message"=>"3"}]
+[3] stdin: [1681358780.517074758, {"message"=>"4"}]
+[4] stdin: [1681358780.517077392, {"message"=>"5"}]
+$
+```
+
+In most cases you'll want to use a more realistic parser that splits your
```
patrick-stephens, line 112: Try not to use shortenings of words so `you'll` should be `you will`. I would also try to avoid any kind of personal pronoun like `you`.

pipeline/inputs/standard-input.md
ringerc: @patrick-stephens The output here isn't json because the example uses the default format of the `stdout` plugin. The _input_ to the `stdin` plugin is required to be valid json unless a parser is configured explicitly.

pipeline/inputs/standard-input.md
ringerc: Reworded to

> In real-world deployments it is best to use a more realistic parser that splits               
> messages into real fields and adds appropriate tags.

pipeline/inputs/standard-input.md
ringerc: @patrick-stephens I can see how it would've been confusing.

* here's what you should provide
* here's the error you get if you violate that expectation

I'll rearrange.

ringerc: Rephrased to

---

If no parser is configured for the stdin plugin, it expects *valid json* input input data with one of the following data formats:
 
```bash
1. { map => val, map => val, map => val }
2. [ time, { map => val, map => val, map => val } ]                                           
```

Any input data that is *not* in one of the above formats will cause the plugin to log errors like: 
   
```
[error] [input:stdin:stdin.0] invalid record found, it's not a JSON map or array
```
 
To handle inputs in other formats, a parser must be explicitly specified in the configuration for the `stdin` plugin. See [parser input example](#parser-input-example) for sample configuration.

ringerc:
@patrick-stephens I've updated this docs PR. I addressed your comments re clarity. 

I've also updated the `concepts/key-concepts.md` doc to reflect the updates made in 9aa10d2bdc0913f94dd366f24b64ef539f0d4b20 since that PR did not update the docs. I wanted to link to the fluent-bit event input format from the docs for the stdin plugin, but found that those docs were outdated so I had to update them.

concepts/key-concepts.md
```diff
@@ -30,12 +30,36 @@ Jan 18 12:52:16 flb gsd-media-keys[2640]: # watch_fast: "/org/gnome/terminal/leg
...
-Internally, an Event always has two components \(in an array form\):
+Internally an Event is comprised of:
+
+* timestamp
+* key/value metadata (since v2.1.0)
+* payload
+
+### Event format
+
+The Fluent Bit wire protocol represents an Event is a 2-element array
```
patrick-stephens, line 41:
```suggestion
The Fluent Bit wire protocol represents an Event as a 2-element array
```

concepts/key-concepts.md
```diff
@@ -30,12 +30,36 @@ Jan 18 12:52:16 flb gsd-media-keys[2640]: # watch_fast: "/org/gnome/terminal/leg
...
+[[TIMESTAMP, METADATA], MESSAGE]
+```
+
+where
+
+* TIMESTAMP is a timestamp in seconds as an integer or floating point value (not a string);
+* METADATA is a possibly-empty object containing event metadata; and
+* MESSAGE is an object containing the event body.
+
+Fluent-bit versions prior to v2.1.0 instead used:
```
patrick-stephens, line 54: My pet peeve, let's be consistent on "Fluent Bit":

```suggestion
Fluent Bit versions prior to v2.1.0 instead used:
```

ringerc:
@patrick-stephens I applied both your changes. Thankyou.

patrick-stephens (approved):

ringerc:
@patrick-stephens Thankyou, and much appreciated.

PR #1099, opened 2023-05-08
pipeline:inputs:forward: Documented new option tag to override the ta…

RicardoAAD:
…g from the forwarded events


RicardoAAD:
Reference to: https://github.com/fluent/fluent-bit/pull/7252

PR #1102, opened 2023-05-12
in_splunk: Add documentation for input splunk plugin (Splunk HTTP HEC)

cosmo0920:
Corresponds to https://github.com/fluent/fluent-bit/pull/7349.

### DIFF
--- a/pipeline/inputs/splunk.md
+++ b/pipeline/inputs/splunk.md
@@ -1,0 +1,56 @@
+# Splunk (HTTP HEC)
+
+The **splunk** input plugin handles [Splunk HTTP HEC](https://docs.splunk.com/Documentation/Splunk/latest/Data/UsetheHTTPEventCollector) requests.
+
+## Configuration Parameters
+
+| **Key**                  | Description                                                                                                                                   | default |
+|--------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|---------|
+| listen                   | The address to listen on                                                                                                                      | 0.0.0.0 |
+| port                     | The port for Fluent Bit to listen on                                                                                                          | 9880    |
+| tag_key                  | Specify the key name to overwrite a tag. If set, the tag will be overwritten by a value of the key.                                           |         |
+| buffer_max_size          | Specify the maximum buffer size in KB to receive a JSON message.                                                                              | 4M      |
+| buffer_chunk_size        | This sets the chunk size for incoming incoming JSON messages. These chunks are then stored/managed in the space available by buffer_max_size. | 512K    |
+| successful_response_code | It allows to set successful response code. `200`, `201` and `204` are supported.                                                              | 201     |
+| splunk\_token           | Add an Splunk token for HTTP HEC.`                                  |         |
+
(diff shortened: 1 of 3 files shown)


pipeline/inputs/splunk.md
```diff
@@ -0,0 +1,56 @@
+# Splunk (HTTP HEC)
+
+The **splunk** input plugin handles Splunk HTTP HEC requests.
```
patrick-stephens, line 3: Probably want to link out to the definition of what these Splunk acronyms are.

pipeline/inputs/splunk.md
```diff
@@ -0,0 +1,56 @@
...
+| successful_response_code | It allows to set successful response code. `200`, `201` and `204` are supported.                                                              | 201     |
+| splunk\_token           | Add an Splunk token for HTTP HEC.`                                  |         |
+
+## Getting Started
+
+In order to start performing the checks, you can run the plugin from the command line or through the configuration file.
+
+#### How to set tag
+
+The tag for the Splunk input plugin is set by adding the tag to the end of the request URL by default.
```
patrick-stephens, line 23: Can you show examples to clarify?

pipeline/inputs/splunk.md
```diff
@@ -0,0 +1,56 @@
...
+This tag is then used to route the event through the system.
+The default behavior of the splunk input sets the tags for the following endpoints:
```
patrick-stephens, line 25: I'm a little confused, it's probably a specific aspect of the Splunk protocol but what does this mean?

We receive an inbound request and the URL provided from that is one of these? We then process it?

pipeline/inputs/splunk.md
```diff
@@ -0,0 +1,56 @@
...
+
+* `/services/collector`
+* `/services/collector/event`
+* `/services/collector/raw`
+
+The requests for these endpoints are interpreted as `services_collector`, `services_collector_event`, and `services_collector_raw`.
+
+If you want to use the other tags for multiple instantiating input splunk plugin, you have to specify `tag` property on the each of splunk plugin configurations to prevent collisions of data pipeline.
```
patrick-stephens, line 33: Again, probably a bit confused and examples will help.

What happens if I set a tag without collision, does it override the defaults or is it ignored (that should be a warning I guess)?

pipeline/inputs/splunk.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Command Line
+
+From the command line you can configure Fluent Bit to handle Bulk API requests with the following options:
```
patrick-stephens, line 37: Not sure why this is here, it doesn't look like we do anything to handle bulk requests. Is this a typo and it should be about just handling Splunk input or is there a differentiation between non-bulk and bulk - how do we enable either in that case?

pipeline/inputs/splunk.md
cosmo0920: The latter lines show the default behavior for handling when not specifying tag property in splunk input plugin.

PR #1116, opened 2023-05-26
s3: always requires write-able filesystem

PettitWesley:


patrick-stephens (approved):
I think this is related to those issues a while back using a read only container filesystem

PR #1135, opened 2023-06-19
in_elasticsearch: Add descriptions for elasticsearch server parameter and default values

cosmo0920:
The corresponding PR is https://github.com/fluent/fluent-bit/pull/7579.

### DIFF
--- a/pipeline/inputs/elasticsearch.md
+++ b/pipeline/inputs/elasticsearch.md
@@ -12,8 +12,10 @@ The plugin supports the following configuration parameters:
 | buffer\_chunk\_size | Set the buffer chunk size. | 512K |
 | tag\_key | Specify a key name for extracting as a tag. | `NULL` |
 | meta\_key | Specify a key name for meta information. | "@meta" |
-| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. | "localhost" |
+| hostname | Specify hostname or FQDN. This parameter is effective for sniffing node information. | "localhost" |
 | version  | Specify Elasticsearch server version. This parameter is effective for checking a version of Elasticsearch/OpenSearch server version. | "8.0.0" |
+
+**Note:** Elasticsearch cluster uses sniffing to optimize the connections between its cluster and clients. Elasticsearch can build its cluster and dynamically generating for connection list is called as sniffing. `hostname` will be reflected for sniffing information which is handled by the sniffing endpoint.
 
 ## Getting Started
 
--- a/pipeline/inputs/elasticsearch.md
+++ b/pipeline/inputs/elasticsearch.md
@@ -6,14 +6,14 @@ The **elasticsearch** input plugin handles both Elasticsearch and OpenSearch Bul
 
 The plugin supports the following configuration parameters:
 
-| Key | Description |
-| :--- | :--- |
-| buffer\_max\_size | Set the maximum size of buffer. |
-| buffer\_chunk\_size | Set the buffer chunk size. |
-| tag\_key | Specify a key name for extracting as a tag. |
-| meta\_key | Specify a key name for meta information. |
-| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. |
-| version  | Specify Elasticsearch server version. This parameter is effective for checking a version of Elasticsearch/OpenSearch server version. |
+| Key | Description | Default value |
+| :--- | :--- | :--- |
+| buffer\_max\_size | Set the maximum size of buffer. | 4M |
+| buffer\_chunk\_size | Set the buffer chunk size. | 512K |
+| tag\_key | Specify a key name for extracting as a tag. | `NULL` |
+| meta\_key | Specify a key name for meta information. | "@meta" |
+| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. | "localhost" |
+| version  | Specify Elasticsearch server version. This parameter is effective for checking a version of Elasticsearch/OpenSearch server version. | "8.0.0" |
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


pipeline/inputs/elasticsearch.md
```diff
@@ -6,13 +6,14 @@ The **elasticsearch** input plugin handles both Elasticsearch and OpenSearch Bul
...
-| tag\_key | Specify a key name for extracting as a tag. |
-| meta\_key | Specify a key name for meta information. |
-| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. |
+| Key | Description | Default value |
+| :--- | :--- | :--- |
+| buffer\_max\_size | Set the maximum size of buffer. | 4M |
+| buffer\_chunk\_size | Set the buffer chunk size. | 512K |
+| tag\_key | Specify a key name for extracting as a tag. | `NULL` |
+| meta\_key | Specify a key name for meta information. | "@meta" |
+| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. | "localhost" |
```
patrick-stephens, line 15: Typo on `sniffering` but I'm not sure what it is supposed to mean here - do we get some extra metadata from it?

patrick-stephens (changes requested):

pipeline/inputs/elasticsearch.md
cosmo0920: Ah, in Elasticsearch/OpenSearch context, sniffer means auto discovery of the server nodes.
ref: https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/current/sniffer.html

Should we use another terms and descriptions for this parameter?

patrick-stephens: Ah right so it is auto-discovery of Elastic nodes. What does this parameter actually do, what should you set it to?

I think it should be the hostname of an Elastic node right? Then the library uses the `info` endpoint to discover the others? Say you're running in a multi-nic of container environment, what would this be set to?

cosmo0920: > I think it should be the hostname of an Elastic node right? Then the library uses the `info` endpoint to discover the others?

Yes, this should be set up with the hostname or FQDN and returned through `info` endpoint.

> Say you're running in a multi-nic of container environment, what would this be set to?

Not sure but, in k8s, users should specify the value as `my-svc.my-namespace.svc.cluster.local`, I think.

patrick-stephens: For K8S, if it's in the same namespace you should only need the service name: you only need the namespace if it is in another namespace. The pod name is HOSTNAME too I think.

cosmo0920: I see. I tested with ${HOSTNAME} as a default value for this parameter. Then, fluent-bit handles to return actual hostname correctly. I didn't know that the translation mechanism in config_map for environment variable.
Should we change the default value of the hostname parameter?

patrick-stephens: Not sure, I think it needs to be explicitly a node in the cluster right? We can do if you want, it should be no worse than localhost.

cosmo0920: > Not sure, I think it needs to be explicitly a node in the cluster right? 

Yes, it is. Currently, we need to specify the cluster FQDN if needed explicitly.
For now, I wanted to leave as-is. Because localhost is always existing value of FQDN. 
Also, in Windows, there is not always mDNS support. Specifying "localhost" is always safe for every plaftorms.

patrick-stephens: OK, can you clarify the node discovery aspects then and add that URL?

cosmo0920: I didn't find the actual documentation for describing sniffing.
Instead, I found the blog post which describes its feature: https://www.elastic.co/jp/blog/elasticsearch-sniffing-best-practices-what-when-why-how
Should we use it instead of the previous URL?

pipeline/inputs/elasticsearch.md
```diff
@@ -6,13 +6,16 @@ The **elasticsearch** input plugin handles both Elasticsearch and OpenSearch Bul
...
-| tag\_key | Specify a key name for extracting as a tag. |
-| meta\_key | Specify a key name for meta information. |
-| hostname | Specify hostname or FQDN. This parameter is effective for sniffering node information. |
+| Key | Description | Default value |
+| :--- | :--- | :--- |
+| buffer\_max\_size | Set the maximum size of buffer. | 4M |
+| buffer\_chunk\_size | Set the buffer chunk size. | 512K |
+| tag\_key | Specify a key name for extracting as a tag. | `NULL` |
+| meta\_key | Specify a key name for meta information. | "@meta" |
+| hostname | Specify hostname or FQDN. This parameter is effective for sniffing node information. | "localhost" |
```
patrick-stephens, line 15:
```suggestion
| hostname | Specify hostname or FQDN. This parameter can be used for "sniffing" (auto-discovery of) cluster node information. | "localhost" |
```

pipeline/inputs/elasticsearch.md
```diff
@@ -6,13 +6,16 @@ The **elasticsearch** input plugin handles both Elasticsearch and OpenSearch Bul
...
+| version  | Specify Elasticsearch server version. This parameter is effective for checking a version of Elasticsearch/OpenSearch server version. | "8.0.0" |
+
+**Note:** Elasticsearch cluster uses sniffing to optimize the connections between its cluster and clients. Elasticsearch can build its cluster and dynamically generating for connection list is called as sniffing. `hostname` will be reflected for sniffing information which is handled by the sniffing endpoint.
```
patrick-stephens, line 18:
```suggestion
**Note:** The Elasticsearch cluster uses "sniffing" to optimize the connections between its cluster and clients.
Elasticsearch can build its cluster and dynamically generate a connection list which is called "sniffing".
The `hostname` will be used for sniffing information and this is handled by the sniffing endpoint.
```

patrick-stephens (approved):

PR #1147, opened 2023-07-03
raspbian11: Use signed-by style on GPG key for apt repo

cosmo0920:


installation/linux/raspbian-raspberry-pi.md
```diff
@@ -5,14 +5,20 @@ Fluent Bit is distributed as **fluent-bit** package and is available for the Ras
 * Raspbian Bullseye \(11\)
 * Raspbian Buster \(10\)
 
-## Server GPG key
+## Server GPG key (old style)
```
patrick-stephens, line 8: I'd say this is confusing, let's just do it one way - the old docs are always around if people want the old way.

patrick-stephens (commented):
Let's only have one way to do it.

Could you update the other Debian targets as well @cosmo0920 to be consistent?
@ehienabs can we have a common section inserted rather than duplicating it everywhere?

installation/linux/raspbian-raspberry-pi.md
cosmo0920: Yup. I understand. Personally, I wanted to add common section to handle apt-key deprecation.

eschabell:
Been doing some research here @cosmo0920 and I'm going to just close this as your branch is effectively obsolete (from 2023):

1. The file was moved in September 2025 from installation/linux/ to installation/downloads/linux/ — the branch is targeting a path that no longer exists in master.
2. The file has been significantly rewritten in master. The current version at the new path is much cleaner.
3. The signed-by change is NOT in the new file — Raspbian 11 (Bullseye) at line 32 still uses a plain deb line without signed-by. So the original intent of the branch is still a valid fix, but it needs to be applied to the new file location.

I'm going to open a fresh PR for this and target the newer situation for the file while marking this `STALE`.

PR #1180, opened 2023-08-22
out_vivo_exporter:  Correct vivo exporter examples

drbugfinder-work:
Examples were missing `match` property


edsiper:
thank you

PR #1181, opened 2023-08-23
Document missing s3 profile option.

jmcarp:
Looks like this option was added a few months ago but not documented.


lecaros:
The profile option was already documented.

PR #1189, opened 2023-09-05
in_kubernetes_events: Fix sample

kaitoii11:
Fix the text and sample configuration for Kubernetes Events.

I went ahead and fixed the sample configuration to align with the description. However, I wasn't sure if the Kubernetes events work with Prometheus exporters?
If not I can fix it to use stdout instead.


kaitoii11:
It seems like someone else has fixed the issue.

PR #1190, opened 2023-09-06
Complete the documentation of the legacy mode

tguenneguez:


tguenneguez:
Some reviewer ?

lecaros:
@nokute78 can you please validate this is accurate? I saw some code of yours in the grep filter :)

nokute78:
@lecaros This patch is for `legacy` mode and it was implemented by @edsiper 

I sent a following patch to support `legacy`.
`legacy` means a first implementation of filter_grep.
https://github.com/fluent/fluent-bit/pull/6721

PR #1191, opened 2023-09-06
docs(opensearch): record accessor syntax links

hown3d:
Link to record accessor docs was wrong.
Syntax in FAQ example used double quotes instead of single quotes, which is not valid.


lecaros (approved):

lecaros:
Thanks @hown3d! Merged.

PR #1198, opened 2023-09-20
in_process_exporter_metrics: Add documentation for process_exporter_metrics

cosmo0920:
Corresponding to https://github.com/fluent/fluent-bit/pull/7943.

### DIFF
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -82,6 +82,6 @@ docker run -ti -v /proc:/host/proc \
 
 ## Enhancement Requests
 
-Our current plugin refers a sub-set of the available collectors in the 3rd party implementation of Prometheus Process Exporter, if you would like that we prioritize a specific metrics please open a Github issue by using the following template:\
+Development prioritises a subset of the available collectors in the [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter), to request others please open a Github issue by using the following template:\
 \
 \- [in_process_exporter_metrics](https://github.com/fluent/fluent-bit/issues/new?assignees=\&labels=\&template=feature_request.md\&title=in_process_exporter_metrics:%20add%20ABC%20collector)
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -9,7 +9,7 @@ description: >-
 [Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics.
 Fluent Bit 2.2 onwards includes a process exporter plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
 
-The initial release of Process Exporter Metrics refers a process group level collector to collect metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and we plan to expand them over time if needed.
+The Process Exporter Metrics plugin implements collecting of the various metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and these will be expanded over time as needed.
 
 **Important note:** Metrics collected with Process Exporter Metrics flow through a separate pipeline from logs and current filters do not operate on top of metrics.
 
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -57,7 +57,7 @@ In the following configuration file, the input plugin _process\_exporter\_metric
     port            2021
 ```
 
-You can test the expose of the metrics by using _curl:_
+You can see the metrics by using _curl:_
 
 ```bash
 curl http://127.0.0.1:2021/metrics
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -23,7 +23,7 @@ macOS does not have procfs. Therefore, this plugin won't work for it.
 | Key                       | Description                                                                            | Default   |
 | ------------------------- | -------------------------------------------------------------------------------------- | --------- |
 | scrape_interval           | The rate at which metrics are collected.                 | 5 seconds |
-| path.procfs               | The mount point used to collect process information and metrics                        | /proc/    |
+| path.procfs               | The mount point used to collect process information and metrics. Read-only is enough   | /proc/    |
 | process\_include\_pattern | regex to determine which processes are included in the metrics produced by this plugin | It is applied for all process unless explicitly set. Default is `.+`. |
 | process\_exclude\_pattern | regex to determine which units are excluded in the metrics produced by this plugin     | It is not applied unless explicitly set. Default is `NULL`. |
 
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -6,7 +6,8 @@ description: >-
 
 # Process Exporter Metrics
 
-[Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics. Fluent Bit 2.2 includes process exporter metrics plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
+[Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics.
+Fluent Bit 2.2 onwards includes a process exporter plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
 
 The initial release of Process Exporter Metrics refers a process group level collector to collect metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and we plan to expand them over time if needed.
 
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -24,8 +24,8 @@ macOS does not have procfs. Therefore, this plugin won't work for it.
 | ------------------------- | -------------------------------------------------------------------------------------- | --------- |
 | scrape_interval           | The rate at which metrics are collected.                 | 5 seconds |
 | path.procfs               | The mount point used to collect process information and metrics. Read-only is enough   | /proc/    |
-| process\_include\_pattern | regex to determine which processes are included in the metrics produced by this plugin | It is applied for all process unless explicitly set. Default is `.+`. |
-| process\_exclude\_pattern | regex to determine which units are excluded in the metrics produced by this plugin     | It is not applied unless explicitly set. Default is `NULL`. |
+| process\_include\_pattern | regex to determine which names of processes are included in the metrics produced by this plugin | It is applied for all process unless explicitly set. Default is `.+`. |
+| process\_exclude\_pattern | regex to determine which names of processes are excluded in the metrics produced by this plugin | It is not applied unless explicitly set. Default is `NULL`. |
 
 ## Getting Started
 
--- a/pipeline/inputs/process-exporter-metrics.md
+++ b/pipeline/inputs/process-exporter-metrics.md
@@ -13,7 +13,7 @@ The Process Exporter Metrics plugin implements collecting of the various metrics
 
 **Important note:** All metrics including those collected with this plugin flow through a separate pipeline from logs and current filters do not operate on top of metrics.
 
-This plugin is only supported on Linux based operating systems due to heavily relying on proc filesystem.
+This plugin is only supported on Linux based operating systems as it uses the `proc` filesystem to access the relevant metrics.
 
 macOS does not have procfs. Therefore, this plugin won't work for it.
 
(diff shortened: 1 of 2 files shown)


pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
+---
+description: >-
+  A plugin based on Process Exporter to collect process level of metrics of system
+  metrics
+---
+
+# Process Exporter Metrics
+
+[Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics. Fluent Bit 2.2 includes process exporter metrics plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
```
patrick-stephens, line 9: Let's split into a sentence per line for reviewing.
```suggestion
[Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics.
Fluent Bit 2.2 onwards includes a process exporter plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
```

Explain what process level vs system level is.

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+The initial release of Process Exporter Metrics refers a process group level collector to collect metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and we plan to expand them over time if needed.
```
patrick-stephens, line 11:
```suggestion
The Process Exporter Metrics plugin implements collecting of the various metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and these will be expanded over time as needed.
```

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Important note:** Metrics collected with Process Exporter Metrics flow through a separate pipeline from logs and current filters do not operate on top of metrics.
```
patrick-stephens, line 13: This is true for all metrics plugins - we really should have a common page for it then just link out.

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+This plugin is only supported on Linux based operating systems.
```
patrick-stephens, line 15: What about macOS?

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Configuration
+
+| Key                       | Description                                                                            | Default   |
+| ------------------------- | -------------------------------------------------------------------------------------- | --------- |
+| scrape_interval           | The rate at which metrics are collected from the host operating system                 | 5 seconds |
```
patrick-stephens, line 22:
```suggestion
| scrape_interval           | The rate at which metrics are collected.                 | 5 seconds |
```

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+| path.procfs               | The mount point used to collect process information and metrics                        | /proc/    |
```
patrick-stephens, line 23: Any specification on the type of mount required, e.g. is read-only ok?

patrick-stephens: Probably want an example for containers here - we cannot mount `/proc` so have to customise.

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+| process\_include\_pattern | regex to determine which processes are included in the metrics produced by this plugin | It is applied for all process unless explicitly set. Default is `.+`. |
+| process\_exclude\_pattern | regex to determine which units are excluded in the metrics produced by this plugin     | It is not applied unless explicitly set. Default is `NULL`. |
```
patrick-stephens, line 25: I think these two descriptions slightly confuse things.

What is the regex applied to, the process name? What if you run multiple processes of the same name, how do you distinguish them?

The exclude pattern talks about `units` - are these systemd units? What about non-systemd deployments in that case (e.g. containers)?

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+You can test the expose of the metrics by using _curl:_
+
+```bash
+curl http://127.0.0.1:2021/metrics
+```
+
+### Container to Collect Host Metrics
+
+When deploying Fluent Bit in a container you will need to specify additional settings to ensure that Fluent Bit has access to the host operating system. The following docker command deploys Fluent Bit with specific mount path for ptocfs and settings enabled to ensure that Fluent Bit can collect from the host. These are then exposed over port 2021.
```
patrick-stephens, line 67: Separate lines for review.
```suggestion
When deploying Fluent Bit in a container you will need to specify additional settings to ensure that Fluent Bit has access to the process details.
The following `docker` command deploys Fluent Bit with a specific mount path for `procfs` and settings enabled to ensure that Fluent Bit can collect from the host. 
These are then exposed over port 2021.
```

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+docker run -ti -v /proc:/host/proc \
+               -v /sys:/host/sys   \
```
patrick-stephens, line 71: Why do we need `/sys`? We have not mentioned it anywhere...

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+               -p 2021:2021        \
+               fluent/fluent-bit:2.2.0 \
```
patrick-stephens, line 73:
```suggestion
               fluent/fluent-bit:2.2 \
```

pipeline/inputs/process-exporter-metrics.md
cosmo0920: macOS does not have procfs. So, we can support it only for Linux-based operating systems.

pipeline/inputs/process-exporter-metrics.md
cosmo0920: Oh, I forgot to remove this...

pipeline/inputs/process-exporter-metrics.md
patrick-stephens: Can we add a comment to that effect? I would consider macOS to be in the "Linux-based" terminology so it is good to be clear it is not supported.

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,84 @@
...
+```bash
+curl http://127.0.0.1:2021/metrics
+```
+
+### Container to Collect Host Metrics
+
+When deploying Fluent Bit in a container you will need to specify additional settings to ensure that Fluent Bit has access to the host operating system. The following docker command deploys Fluent Bit with specific mount path for ptocfs and settings enabled to ensure that Fluent Bit can collect from the host. These are then exposed over port 2021.
+
+```
+docker run -ti -v /proc:/host/proc \
```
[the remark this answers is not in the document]
cosmo0920: We don't write any contents into procfs. So, we can use readonly option here.

pipeline/inputs/process-exporter-metrics.md
cosmo0920: macOS does not based on Linux-based technology. It *is* based on BSD 4.2 and Mach technology. filesystem is built on top of IOKit which is made by Apple actually....

patrick-stephens: Sure, I understand but I think it's easy to mistake it for "Linux-based" so be good to ensure we clarify this is not supported on macOS. That's my only point, clarify it is not supported because of the lack of `procfs`.

cosmo0920: Sounds reasonable. I added descriptions for procfs and won't supported it for macOS: https://github.com/fluent/fluent-bit-docs/pull/1198/commits/70ad3c2c905d1ce24578873ca5f09eb054e94266

pipeline/inputs/process-exporter-metrics.md
cosmo0920: Right. I added a note for read-only attribute: https://github.com/fluent/fluent-bit-docs/pull/1198/commits/d8e073872e2bb35794ff7847ed2984ce905fabb8

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,88 @@
...
+macOS does not have procfs. Therefore, this plugin won't work for it.
+
+
+## Configuration
+
+| Key                       | Description                                                                            | Default   |
+| ------------------------- | -------------------------------------------------------------------------------------- | --------- |
+| scrape_interval           | The rate at which metrics are collected.                 | 5 seconds |
+| path.procfs               | The mount point used to collect process information and metrics. Read-only is enough   | /proc/    |
+| process\_include\_pattern | regex to determine which names of processes are included in the metrics produced by this plugin | It is applied for all process unless explicitly set. Default is `.+`. |
```
patrick-stephens, line 27: Can we clarify the regex handling?

It would be useful to have examples for how to monitor specific processes.
Is it the binary name itself as reported by `ps` for example or something else (e.g. if symlinked with a different name?).
Can we/how do we filter processes with the same name but different args? 
I presume we cannot filter by pid either?

cosmo0920: 1. Yes. The binary names are handled as ps reported names which is limited 15 characters.
2. No. We cannot filtered by command line arguments like as prometheus process_exporter does.
3. Correct. We cannot filter by pid.

pipeline/inputs/process-exporter-metrics.md
patrick-stephens:
```suggestion
**Important note:** All metrics including those collected with this plugin flow through a separate pipeline from logs and current filters do not operate on top of metrics.
```

pipeline/inputs/process-exporter-metrics.md
patrick-stephens: Thanks @cosmo0920.

Can we add some examples?

Does the max-length include the path to the binary? 15 seems super small...

cosmo0920: Yup. This seems to be super small and I thought it was wrong.... However, the specification is correct.

The limits is originated from: 
* https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L292
* https://elixir.bootlin.com/linux/latest/source/include/linux/sched.h#L1073

And according to man 2 prctl:

> PR_SET_NAME (since Linux 2.6.9)
              Set  the  name  of the calling thread, using the value in the location pointed to by (char *) arg2.  The name can be up to 16 bytes long, including the terminating null byte.  (If the length of
              the string, including the terminating null byte, exceeds 16 bytes, the string is silently truncated.)  This is the same attribute that can be set via pthread_setname_np(3) and  retrieved  using
              pthread_getname_np(3).  The attribute is likewise accessible via /proc/self/task/[tid]/comm, where tid is the name of the calling thread.

patrick-stephens: Can we add these details in a sub-section maybe with some examples?

cosmo0920: Sure, I'll add them

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,103 @@
...
+# Process Exporter Metrics
+
+[Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is a popular way to collect system level metrics from operating systems, such as CPU / Disk / Network / Process statistics.
+Fluent Bit 2.2 onwards includes a process exporter plugin that builds off the Prometheus design to collect process level metrics without having to manage two separate processes or agents.
+
+The Process Exporter Metrics plugin implements collecting of the various metrics available from [the 3rd party implementation of Prometheus Process Exporter](https://github.com/ncabatoff/process-exporter) and these will be expanded over time as needed.
+
+**Important note:** All metrics including those collected with this plugin flow through a separate pipeline from logs and current filters do not operate on top of metrics.
+
+This plugin is only supported on Linux based operating systems due to heavily relying on proc filesystem.
```
patrick-stephens, line 16:
```suggestion
This plugin is only supported on Linux based operating systems as it uses the `proc` filesystem to access the relevant metrics.
```

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,103 @@
...
+
+macOS does not have procfs. Therefore, this plugin won't work for it.
```
patrick-stephens, line 18:
```suggestion
macOS does not have the `proc` filesustem so this plugin will not work for it.
```

pipeline/inputs/process-exporter-metrics.md
```diff
@@ -0,0 +1,103 @@
...
+The following `docker` command deploys Fluent Bit with a specific mount path for `procfs` and settings enabled to ensure that Fluent Bit can collect from the host. 
+These are then exposed over port 2021.
+
+```
+docker run -ti -v /proc:/host/proc:ro \
+               -p 2021:2021        \
+               fluent/fluent-bit:2.2 \
+               /fluent-bit/bin/fluent-bit \
+                         -i process_exporter_metrics -p path.procfs=/host/proc  \
+                         -o prometheus_exporter -p "add_label=host $HOSTNAME" \
```
patrick-stephens, line 95: we need to specify the `HOSTNAME` variable btw (`-e HOSTNAME="$HOSTNAME"`) or just remove it might be easier

cosmo0920: Sure. I removed it from the doc.

patrick-stephens (approved):

cosmo0920:
🙌

PR #1209, opened 2023-09-28
hot-reloading: Add descriptions for Windows

cosmo0920:
Corresponds to https://github.com/fluent/fluent-bit/pull/7984.

### DIFF
--- a/administration/hot-reload.md
+++ b/administration/hot-reload.md
@@ -39,6 +39,12 @@ For using curl to reload fluent-bit, users must specify an empty request body as
 $ curl -X POST -d {} localhost:2020/api/v2/reload
 ```
 
+For Windows, to represent empty body, users have to use `'{}'` as a body instead of raw `{}`:
+
+```text
+PS> curl -X POST -d '{}' localhost:2020/api/v2/reload
+```
+
 ### Via Signal
 
 Hot reloading also can be kicked via `SIGHUP`.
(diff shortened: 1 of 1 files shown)


administration/hot-reload.md
```diff
@@ -39,6 +39,12 @@ For using curl to reload fluent-bit, users must specify an empty request body as
 $ curl -X POST -d {} localhost:2020/api/v2/reload
 ```
 
+For Windows, to represent empty body, users have to use `'{}'` as a body instead of raw `{}`:
```
patrick-stephens, line 42:
```suggestion
On Windows, use `'{}'` to represent an empty body instead of raw `{}`:
```

patrick-stephens: To be fair you could do this on Linux too I think?

cosmo0920: No, it's Windows specific.

patrick-stephens: No, I mean you can quote it on Linux as well. It works fine so maybe we should just use the one format for both?

Both work for Linux:
```shell
$ curl -X POST -d {} 127.0.0.1:2020/api/v2/reload
{"reload":"done","status":0}
$ curl -X POST -d '{}' 127.0.0.1:2020/api/v2/reload
{"reload":"done","status":0}
```

cosmo0920: Ah, OK. We can do that. Let's move to use quoted empty body.

cosmo0920:
@lecaros The corresponding PR is only included in 2.2 dev branch. IMO, we needn't backport to 2.1 branch...

lecaros:
Understood @cosmo0920. I'll remove it from the backport list

PR #1245, opened 2023-10-26
installation: platforms: add s390x to supported architectures

ajaypvictor:
Adding doc changes as part of enabling s390x.

Related PR: https://github.com/fluent/fluent-bit/pull/8095

### DIFF
--- a/installation/supported-platforms.md
+++ b/installation/supported-platforms.md
@@ -14,7 +14,7 @@ The following operating systems and architectures are supported in Fluent Bit.
 |  | [Debian 12 \(Bookworm\)](linux/debian.md) | x86\_64, Arm64v8 |
 |  | [Debian 11 \(Bullseye\)](linux/debian.md) | x86\_64, Arm64v8 |
 |  | [Debian 10 \(Buster\)](linux/debian.md) | x86\_64, Arm64v8 |
-|  | [Ubuntu 22.04 \(Jammy Jellyfish\)](linux/ubuntu.md) | x86\_64, Arm64v8, s390x |
+|  | [Ubuntu 22.04 \(Jammy Jellyfish\)](linux/ubuntu.md) | x86\_64, Arm64v8 |
 |  | [Ubuntu 20.04 \(Focal Fossa\)](linux/ubuntu.md) | x86\_64, Arm64v8 |
 |  | [Ubuntu 18.04 \(Bionic Beaver\)](linux/ubuntu.md) | x86\_64, Arm64v8 |
 |  | [Ubuntu 16.04 \(Xenial Xerus\)](linux/ubuntu.md) | x86\_64 |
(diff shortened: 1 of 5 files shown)


installation/supported-platforms.md
```diff
@@ -29,3 +29,5 @@ From an architecture support perspective, Fluent Bit is fully functional on x86\
 Fluent Bit can work also on OSX and \*BSD systems, but not all plugins will be available on all platforms.
 Official support will be expanding based on community demand.
 Fluent Bit may run on older operating systems though will need to be built from source, or use custom packages from [enterprise providers](https://fluentbit.io/enterprise).
+
+Fluent Bit is supported on Linux on IBM Z (s390x) environment without WASM, because building WAMR is not supported as of today on s390x and LuaJIT as well.
```
patrick-stephens, line 33: I think this needs a little reword and should it be in the dedicated plugins themselves as well/instead of @lecaros ?

```suggestion
Fluent Bit is supported on Linux on IBM Z (s390x) environment with some restrictions but only container images are provided for these targets officially.
```

installation/supported-platforms.md
```diff
@@ -14,7 +14,7 @@ The following operating systems and architectures are supported in Fluent Bit.
 |  | [Debian 12 \(Bookworm\)](linux/debian.md) | x86\_64, Arm64v8 |
 |  | [Debian 11 \(Bullseye\)](linux/debian.md) | x86\_64, Arm64v8 |
 |  | [Debian 10 \(Buster\)](linux/debian.md) | x86\_64, Arm64v8 |
-|  | [Ubuntu 22.04 \(Jammy Jellyfish\)](linux/ubuntu.md) | x86\_64, Arm64v8 |
+|  | [Ubuntu 22.04 \(Jammy Jellyfish\)](linux/ubuntu.md) | x86\_64, Arm64v8, s390x |
```
patrick-stephens, line 17: Is this true? We do not seem to explicitly do anything to provide any DEBs for s390x which is what this table is showing - actual packages are built and provided officially from our repos. Not that it can be but that it is.

The only changes in the linked PR are to build images, not packages, so want to confirm. If we want to build it via the normal `./packaging/build.sh -d <distro>` then how do we do that?

I would extend the statement about container images being provided with something along the lines of "it is possible to build for Ubuntu 22.04 as well".

pipeline/inputs/exec.md
```diff
@@ -8,7 +8,7 @@ command arguments could lead to malicious command execution.
 
 ## Container support
 
-This plugin will not function in the distroless production images (AMD64 currently) as it needs a functional `/bin/sh` which is not present.
+This plugin will not function in the distroless production images (AMD64 and s390x currently) as it needs a functional `/bin/sh` which is not present.
```
patrick-stephens, line 11: This needs updating, all production images are distroless so this was out of date for a while. No container images support this plugin other than the debug ones.

installation/supported-platforms.md
ajaypvictor: removed..

pipeline/inputs/exec.md
ajaypvictor: updated..

patrick-stephens (approved):

PR #1242, opened 2023-10-20
 Fixing typo in input plugins changed TTL to TLS

RicardoAAD:
Fixing this typo on nine input plugins

PR #1262, opened 2023-11-24
pipeline: fix typo TTL/SSL -> TLS/SSL

ecerulm:


ecerulm:
I guess this does the same as #1242, but that one seems stale.

lecaros:
ping @RicardoAAD

PR #1278, opened 2023-12-22
Update stackdriver.md with warning about Workers option

gerry-tock:
Adding this info about this issue because I lost a bunch of time trying to chase down intermittent restarts of my fluent-bit pods (v.2.2.0). 

Even if it defaults to 1 in newer versions, someone may override it, as was my case. 

Until the underlying issue is fixed, I think it's helpful to warn others about this.

### DIFF
--- a/pipeline/outputs/stackdriver.md
+++ b/pipeline/outputs/stackdriver.md
@@ -184,7 +184,17 @@ Do following check:
 * If the log entry does not contain the local_resource_id field, does the tag of the log match for format?
 *   If tag_prefix is configured, does the prefix of tag specified in the input plugin match the tag_prefix?
 
-    **Other implementations**
+### caught signal (SIGSEGV)
+
+> Github reference: [#7552](https://github.com/fluent/fluent-bit/issues/7552)
+
+When the number of Workers is greater than 1, fluent-bit may interimittently crash.
+
+```
+[2023/06/07 08:15:30] [engine] caught signal (SIGSEGV)
+```
+
+## Other implementations
 
 Stackdriver officially supports a [logging agent based on Fluentd](https://cloud.google.com/logging/docs/agent).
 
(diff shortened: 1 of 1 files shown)


pipeline/outputs/stackdriver.md
```diff
@@ -184,7 +184,17 @@ Do following check:
 * If the log entry does not contain the local_resource_id field, does the tag of the log match for format?
 *   If tag_prefix is configured, does the prefix of tag specified in the input plugin match the tag_prefix?
 
-    **Other implementations**
+### caught signal (SIGSEGV)
```
braydonk, line 187: I would recommend changing this heading, because `caught signal (SIGSEGV)` can be any manner of different things. How about:
```suggestion
### Crashing with default number of `Workers`
```

pipeline/outputs/stackdriver.md
```diff
@@ -184,7 +184,17 @@ Do following check:
...
+
+> Github reference: [#7552](https://github.com/fluent/fluent-bit/issues/7552)
+
+When the number of Workers is greater than 1, fluent-bit may interimittently crash.
+
+```
+[2023/06/07 08:15:30] [engine] caught signal (SIGSEGV)
```
braydonk, line 194: Similarly to the title, this could be anything and isn't specific to this crash with the workers so I would recommend not including it here.

pipeline/outputs/stackdriver.md
```diff
@@ -184,7 +184,17 @@ Do following check:
 * If the log entry does not contain the local_resource_id field, does the tag of the log match for format?
 *   If tag_prefix is configured, does the prefix of tag specified in the input plugin match the tag_prefix?
 
-    **Other implementations**
+### caught signal (SIGSEGV)
+
+> Github reference: [#7552](https://github.com/fluent/fluent-bit/issues/7552)
+
+When the number of Workers is greater than 1, fluent-bit may interimittently crash.
```
braydonk, line 191:
```suggestion
When the number of Workers is greater than 1, Fluent Bit may interimittently crash.
```

patrick-stephens:
@gerry-tock could you just tweak based on the comments? I agree it's important to merge.

pipeline/outputs/stackdriver.md
patrick-stephens: Probably say "you may see this" or something similar. However it could be any crash really.

patrick-stephens:
@braydonk if you approve it should auto-merge

braydonk (approved):

PR #1287, opened 2024-01-10
Added documentation for uri parameter to outputs/loki

PeterStolz:
I opened https://github.com/fluent/fluent-bit/issues/8348 thinking that there was no option to set uri/path for the loki output plugin, but when checking how to implement it I found out it already exists

https://github.com/fluent/fluent-bit/blob/cd497e38dc8e4e958756b449aab34801da486ad2/plugins/out_loki/loki.c#L1741-L1748


PeterStolz:
Nevermind. While it exists in source there is some validation inbetween preventing me from using it directly

PeterStolz:
Nevermind I did not run the latest version of fluent-bit. Setting uri works fine.

patrick-stephens (approved):
Verified via `docker run --rm -it fluent/fluent-bit:2.2.1 -o loki --help`

PR #1294, opened 2024-01-31
filter_aws: filter_ecs: Reflect filter enrichment status

cosmo0920:


cosmo0920:
Ah yes. That PR should be going to be handled as processor plugins instead. So, we can close and revisit similar documents for now.

PR #1300, opened 2024-02-03
in_disk: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for in_disk.


lecaros (approved):

lecaros:
Thanks @nokute78! Merged.

PR #1306, opened 2024-02-09
in_node_exporter_metrics: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for in_node_exporter_metrics .


lecaros:
Thanks @nokute78! Merged.

PR #1309, opened 2024-02-10
filter_rewrite_tag: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for filter_rewrite_tag.


lecaros:
Thanks @nokute78! Merged.

PR #1316, opened 2024-02-23
in_http: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for in_http.


lecaros:
Thanks @nokute78! Merged.

PR #1327, opened 2024-03-02
in_netif: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for in_netif.


lecaros:
Thanks @nokute78! Merged.

PR #1329, opened 2024-03-03
in_stdin: add examples for yaml configuration

nokute78:
This patch is to add an example in yaml for in_stdin.


lecaros:
Thanks @nokute78! Merged.

PR #1350, opened 2024-04-05
in_winevtlog: Add a description for Read_Limit_Per_Cycle

cosmo0920:
This PR corresponds to https://github.com/fluent/fluent-bit/pull/8677.

### DIFF
--- a/pipeline/inputs/windows-event-log-winevtlog.md
+++ b/pipeline/inputs/windows-event-log-winevtlog.md
@@ -17,7 +17,7 @@ The plugin supports the following configuration parameters:
 | Render\_Event\_As\_XML | Whether to render system part of event as XML string or not. \(optional\) | False  |
 | Use\_ANSI | Use ANSI encoding on eventlog messages. If you have issues receiving blank strings with old Windows versions (Server 2012 R2), setting this to True may solve the problem. \(optional\) | False  |
 | Event\_Query | Specify XML query for filtering events. | `*` |
-| Threshold\_Size| Specify threshold size per a cycle. (\optional\)  | |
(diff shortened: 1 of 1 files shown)


pipeline/inputs/windows-event-log-winevtlog.md
```diff
@@ -17,6 +17,7 @@ The plugin supports the following configuration parameters:
 | Render\_Event\_As\_XML | Whether to render system part of event as XML string or not. \(optional\) | False  |
 | Use\_ANSI | Use ANSI encoding on eventlog messages. If you have issues receiving blank strings with old Windows versions (Server 2012 R2), setting this to True may solve the problem. \(optional\) | False  |
 | Event\_Query | Specify XML query for filtering events. | `*` |
+| Read\_Limit\_Per\_Cycle | Specify reading limitation per a cycle. (\optional\)  | |
```
patrick-stephens, line 20:
```suggestion
| Read\_Limit\_Per\_Cycle | Specify read limit per cycle.  | 512KiB |
```

Probably need to define some of the terms and clarify any defaults.
What's a cycle?
What's the units?
What happens if you don't? No limit? Looks like it defaults to 512KiB

cosmo0920: Its unit is bytes. 
A cycle means that the iteration/loop for collecting Windows EventLog. This should be needed because we need to create adequate size of chunks. Normally, it's around 2MiB.
So, it's working like cutoff even if remaining the uncollected Windows EventLogs.
The limit is now specified as 512KiB.

patrick-stephens (approved):

PR #1370, opened 2024-05-13
added docs for Microsoft Fabric Real Time Analytics

tanmaya-panda1:
Microsoft Fabric Real Time Analytics is a SAAS offering in Microsoft Fabric while Azure Kusto is a PAAS offering in Azure. Microsoft Fabric Real Time Analytics has eventhouse(Kusto) for data storage. Hence we are using the same output plugin we have for Azure Kusto to send log data to Microsoft Fabric Real Time Analytics. 

For more details please refer to the following documentations
https://learn.microsoft.com/en-us/fabric/real-time-analytics/overview
https://learn.microsoft.com/en-us/fabric/real-time-analytics/realtime-analytics-compare

### DIFF
--- a/pipeline/outputs/microsoft-fabric-rta.md
+++ b/pipeline/outputs/microsoft-fabric-rta.md
@@ -1,0 +1,83 @@
+---
+description: Send logs to Eventhouse(Kusto) in Microsoft Fabric Real Time Analytics
+---
+
+# Microsoft Fabric Real Time Analytics
+
+This output plugin allows to ingest your logs into an [Eventhouse](https://blog.fabric.microsoft.com/en-us/blog/eventhouse-overview-handling-real-time-data-with-microsoft-fabric/) cluster in Microsoft Fabric Real Time Analytics, via the [Queued Ingestion](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/about-kusto-ingest#queued-ingestion) mechanism.
+
+## Creating an Eventhouse Cluster and KQL Database
+
+You can create an Eventhouse cluster and a KQL database in one of the following ways:
+
+- [Create a Eventhouse cluster](https://learn.microsoft.com/en-us/fabric/real-time-analytics/create-eventhouse)
+- [Create a KQL database in Eventhouse](https://learn.microsoft.com/en-us/fabric/real-time-analytics/create-database)
+
+## Creating an Azure Registered Application
+
+Fluent-Bit will use the application's credentials, to ingest data into your cluster.
+
+- [Register an Application](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app#register-an-application)
+- [Add a client secret](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app#add-a-client-secret)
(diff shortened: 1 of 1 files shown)


patrick-stephens:
Please link to the PR providing this functionality

tanmaya-panda1:
@patrick-stephens , this functionality is already existing in fluent-bit output plugin in the name of azure_kusto. Kusto is a PAAS offering in Azure but Microsoft has introduced a new SAAS offering named Microsoft Fabric Real Time Analytics whose core is Kusto. So we use the same output plugin to send data to Microsoft Fabric.

patrick-stephens:
Is this a duplicate of the other doc then? That's not ideal as they will change over time - can we link them instead and just have a separate section in the original?

kustonaut:
@patrick-stephens - To help Fluent bit users discover the output plugin to Microsoft Fabric, we wanted to list the same on this fluent bit output page: https://docs.fluentbit.io/manual/pipeline/outputs . Could you please guide us how we can achieve the same? The same output plugin works for both Azure Data Explorer and Microsoft Fabric.
image

patrick-stephens:
> @patrick-stephens - To help Fluent bit users discover the output plugin to Microsoft Fabric, we wanted to list the same on this fluent bit output page: https://docs.fluentbit.io/manual/pipeline/outputs . Could you please guide us how we can achieve the same? The same output plugin works for both Azure Data Explorer and Microsoft Fabric. 

Right there's two things here.

1. I think we should update the original Azure output to include the relevant details that it works for this fancy new named thing too.

2. Separately we can also add a link to that same page with the new name via https://github.com/fluent/fluent-bit-docs/blob/master/SUMMARY.md

Is there a reason it is called different names or is it just rebranding? My understanding is that it's just a PAAS offering so I'm not overly keen on adding vendor marketing to the docs but I can understand users may want it to help differentiate. Similar to what we do for Elastic output also referencing Elastic Cloud.

kustonaut:
> > @patrick-stephens - To help Fluent bit users discover the output plugin to Microsoft Fabric, we wanted to list the same on this fluent bit output page: https://docs.fluentbit.io/manual/pipeline/outputs . Could you please guide us how we can achieve the same? The same output plugin works for both Azure Data Explorer and Microsoft Fabric.
> 
> Right there's two things here.
> 
> 1. I think we should update the original Azure output to include the relevant details that it works for this fancy new named thing too.
> 2. Separately we can also add a link to that same page with the new name via https://github.com/fluent/fluent-bit-docs/blob/master/SUMMARY.md
> 
> Is there a reason it is called different names or is it just rebranding? My understanding is that it's just a PAAS offering so I'm not overly keen on adding vendor marketing to the docs but I can understand users may want it to help differentiate. Similar to what we do for Elastic output also referencing Elastic Cloud.

Microsoft Fabric is a combination of multiple Microsoft services like Power BI, Datawarehouse, Data Factory etc. ADX/Kusto forms of the pillars of Microsoft Fabric and is named as Real-Time Analytics. Fabric users who want to bring telemetry using fluentbit might not be aware of ADX/Kusto and might end up confused if we point them to existing documentation where these terms are used. Kusto/ADX is just the underlying engine there but the experience is very different from the PaaS version.

tanmaya-panda1:
@patrick-stephens Thank you for your inputs, we have created a separate PR https://github.com/fluent/fluent-bit-docs/pull/1372 with the changes you have suggested. Can you please review the same.

PR #1380, opened 2024-05-24
out_s3: add description of parquet compression

cosmo0920:
Corresponding to https://github.com/fluent/fluent-bit/pull/8837.


pipeline/outputs/s3.md
```diff
@@ -40,7 +40,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
... compression                      | Compression type for S3 objects. 'gzip' and 'parquet' are currently the only supported value by default. If Apache Arrow support was enabled at compile time, you can also use 'arrow'. If columnify command is installed, you can also compress as parquet format. For gzip compression, the Content-Encoding HTTP Header will be set to 'gzip'. Gzip and parquet compression can be enabled when `use_put_object` is 'on' or 'off' (PutObject and Multipart). Arrow compression can only be enabled with `use_put_object On`.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  | None                                    |
```
patrick-stephens, line 43: Is `none` a supported type or is it empty?

What happens if you do not set the `use_put_object` option appropriately - does it report a config failure (good) or just fail silently to do what you requested (bad)?

pipeline/outputs/s3.md
```diff
@@ -49,6 +49,13 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| parquet.compression             | Compression type for parquet. 'uncompressed', 'snappy', 'gzip', 'zstd' are the supported values by default. 'lzo', 'brotli', 'lz4' are not supported for now.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          | SNAPPY                                  |
```
patrick-stephens, line 52: lowercase `snappy`?

pipeline/outputs/s3.md
```diff
@@ -49,6 +49,13 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| parquet.pagesize                | Page size of parquet format. Defaults to 8192 bytes (8KiB).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            | 8192                                    |
```
patrick-stephens, line 53: Size in bytes right? Let's update that in these fields to be clear.

pipeline/outputs/s3.md
```diff
@@ -49,6 +49,13 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
+| parquet.schema\_file            | Speficy path to schema file for parquet compression.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   | None                                    |
```
patrick-stephens, line 57: Typo - `specify`

pipeline/outputs/s3.md
```diff
@@ -282,6 +289,23 @@ Example:
 
 Then, the records will be stored into the MinIO server.
 
+## Usage for Parquet Compression
+
+For parquet compression, it needs to install [columnify](https://github.com/reproio/columnify) at runtime.
```
patrick-stephens, line 294: What does this mean? Is it in the image or part of the package? Or do we have to do it separately?

cosmo0920: It means that columnify command should be installed in the system or container at runtime. This is not building dependency but for runtime dependency.

pipeline/outputs/s3.md
cosmo0920: No. Columnify does not accept lower cases of compression methods.

pipeline/outputs/s3.md
cosmo0920: I supposed that it means empty. But all of the documentation of out_s3 uses this habit of usage for None.

pipeline/outputs/s3.md
cosmo0920: Yup, size should be byte(s) unit. However, page size is a proper noun to specify size of "pagesize" for parquet compression.
ref: https://parquet.apache.org/docs/file-format/configurations/

cosmo0920: So, do we need to use _Pagesize_ instead of _Page size_ here?

pipeline/outputs/s3.md
cosmo0920: > What happens if you do not set the use_put_object option appropriately - does it report a config failure (good) or just fail silently to do what you requested (bad)?

This cause just configuration error:
https://github.com/fluent/fluent-bit/pull/8837/files#diff-33a08a0c450f5b99e5c77cd5a5178e09b5482f262721994256817c03ebd0cdfbR687

pipeline/outputs/s3.md
cosmo0920: Added in https://github.com/fluent/fluent-bit-docs/pull/1380/commits/0b6238b03597107d8727c82557635a77a57e03db.

pipeline/outputs/s3.md
patrick-stephens: OK then uppercase the values in the description and let's add a note saying it must be uppercase. Otherwise people will assume you can use whatever like the rest of Fluent Bit config.

pipeline/outputs/s3.md
patrick-stephens: Sorry I meant just make it clear the value you enter is in bytes - someone might put in 8Mb or something silly.

pipeline/outputs/s3.md
patrick-stephens, line 43:
```suggestion
| compression                      | Compression type for S3 objects. 'gzip' and 'parquet' are currently the only supported value by default. If Apache Arrow support was enabled at compile time, you can also use 'arrow'. If columnify command is installed, you can also compress as parquet format. For gzip compression, the Content-Encoding HTTP Header will be set to 'gzip'. Gzip and parquet compression can be enabled when `use_put_object` is 'on' or 'off' (PutObject and Multipart). Arrow compression can only be enabled with `use_put_object On`. A configuration error will be triggered for invalid combinations. Default is empty and no compression therefore.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |                                    |
```

pipeline/outputs/s3.md
cosmo0920: Ah, got it. I'd confused what you mean. Other parameters which are related of byte(s) unit use bytes in the document. So, I added them.

pipeline/outputs/s3.md
cosmo0920: I noticed that the converter which converts from lower case to upper case is already implemented in out_s3. So, I added descriptions for both of lower case and upper case should be accepted this parameter.

patrick-stephens (approved):

pipeline/outputs/s3.md
```diff
@@ -282,6 +289,23 @@ Example:
 
 Then, the records will be stored into the MinIO server.
 
+## Usage for Parquet Compression
+
+For parquet compression, it needs to install [columnify](https://github.com/reproio/columnify) in the running system or container at runtime.
```
PettitWesley, line 294: I'd say, "users need to install"

Can we flesh this out into an example with the FLuent Bit image so folks can easily figure it out?

From reaing this, I don't know how to use parquet quickly

PettitWesley, line 294: I'd say, "users need to install"

Can we flesh this out into an example with the FLuent Bit image so folks can easily figure it out?

From reaing this, I don't know how to use parquet quickly

pipeline/outputs/s3.md
```diff
@@ -40,7 +40,7 @@ See [here](https://github.com/fluent/fluent-bit-docs/tree/43c4fe134611da471e706b
...
... only supported value by default. If Apache Arrow support was enabled at compile time, you can also use 'arrow'. If columnify command is installed, you can also compress as parquet format. For gzip compression, the Content-Encoding HTTP Header will be set to 'gzip'. Gzip and parquet compression can be enabled when `use_put_object` is 'on' or 'off' (PutObject and Multipart). Arrow compression can only be enabled with `use_put_object On`. A configuration error will be triggered for invalid combinations. Default is empty and no compression therefore.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |                                    |
```
PettitWesley, line 43: > columnify

I don't know what this is, can it be a link ?

PettitWesley (changes requested):

pipeline/outputs/s3.md
cosmo0920: Good point. I'll add it with a link.

pipeline/outputs/s3.md
cosmo0920: We'll plan to add an instruction of columnify command into Fluent Bit image at least.

cosmo0920: Not sure where we should put the update for using columnify command but I found that a minimal example to add columnify command in a Docker container:

```docker
FROM golang:1.22-alpine as builder

ENV ROOT=/go/src/cmd
WORKDIR ${ROOT}

RUN apk update && \
  apk add git

RUN go install github.com/reproio/columnify/cmd/columnify@latest

# hadolint ignore=DL3006
FROM debian:bullseye-slim as production

COPY --from=builder /go/bin/columnify /usr/bin/columnify
```

cosmo0920: I tested both of 1.22-bullseye and 1.22-alpine images. When using alpine images, I got succeeded to obtain non-dynamic executables for columnify.

PettitWesley: I think either this or something like it needs to be in the docs, or linked to from the docs.

cosmo0920: I added an example for building columnify command inside containers here: https://github.com/fluent/fluent-bit-docs/pull/1380/commits/94838f393644cce15c6883bc63a064661b33496f

pipeline/outputs/s3.md
```diff
@@ -282,6 +290,44 @@ Example:
...
+
+RUN apk update && \
+  apk add git
+
+RUN go install github.com/reproio/columnify/cmd/columnify@latest
+
+FROM debian:bullseye-slim as production
+
+# Put columnify command inside the PATH.
+COPY --from=builder /go/bin/columnify /usr/bin/columnify
```
PettitWesley, line 328: Does the Dockerfile need a CMD or ENTRYPOINT?

PettitWesley (approved):

pipeline/outputs/s3.md
cosmo0920: I think that it's not needed because it's a just example for building for columnify command.

PR #1388, opened 2024-06-04
Add Time_System_Timezone parser option docs

braydonk:
This PR adds documentation for the proposed `Time_System_Timezone` configuration option.

Related PR: https://github.com/fluent/fluent-bit/pull/8164


patrick-stephens (approved):

patrick-stephens:
@braydonk I presume we don't hit the usual issue with timezones not set in container images? What's the process it uses to get the default timezone (e.g. across OS/containers)?

braydonk:
Under the hood the feature uses the `mktime` function from libc, which under the hood uses the POSIX `TZ` environment variable. This should work on all platforms Fluent Bit supports (including Windows in my testing), but I'm not sure about containers. I don't know if there is anything that Fluent Bit should be expected to do if the `TZ` env var isn't set properly though.

Should I add this to the table entry here in the docs? Specifically I'd mention the `TZ` environment variable.

patrick-stephens:
Yeah, it's a known issue sometimes that TZ may not be set or imported from the host, etc.

patrick-stephens (approved):

braydonk:
The feature was merged in https://github.com/fluent/fluent-bit/pull/9006. This should be good to be merged now.

patrick-stephens:
Is it in a release @braydonk, that's the normal gate rather than just a merge even though the label suggests otherwise :)

PR #1393, opened 2024-06-17
tls: Add descriptions for x509v3 alt name verification (tls.verify_hostname)

cosmo0920:
Corresponding to https://github.com/fluent/fluent-bit/pull/8966.

### DIFF
--- a/administration/transport-security.md
+++ b/administration/transport-security.md
@@ -9,6 +9,7 @@ Both input and output plugins that perform Network I/O can optionally enable TLS
 | :--- | :--- | :--- |
 | tls | enable or disable TLS support | Off |
 | tls.verify | force certificate validation | On |
+| tls.verify\_hostname | force hostname validation for certificates | Off |
 | tls.debug | Set TLS debug verbosity level. It accept the following values: 0 \(No debug\), 1 \(Error\), 2 \(State change\), 3 \(Informational\) and 4 Verbose | 1 |
 | tls.ca\_file | absolute path to CA certificate file |  |
 | tls.ca\_path | absolute path to scan for certificate files |  |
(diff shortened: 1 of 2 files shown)


administration/transport-security.md
```diff
@@ -9,6 +9,7 @@ Both input and output plugins that perform Network I/O can optionally enable TLS
 | :--- | :--- | :--- |
 | tls | enable or disable TLS support | Off |
 | tls.verify | force certificate validation | On |
+| tls.verify\_hostname | force hostname validation for certificates | Off |
```
patrick-stephens, line 12: Slightly different wording to below.

administration/transport-security.md
```diff
@@ -171,3 +172,41 @@ Fluent Bit supports [TLS server name indication](https://en.wikipedia.org/wiki/S
     tls.ca_file /etc/certs/fluent.crt
     tls.vhost   fluent.example.com
 ```
+
+### Verify subjectAltName
+
+When extracting the X509v3 Subject Alternative Name from a certificate:
```
patrick-stephens, line 178: I maybe would reword this a bit, using `Then` as well is not ideal.

Something like?
```suggestion.
By default, TLS verification of hostnames is not done automatically.
As an example, we can extract the X509v3 Subject Alternative Name from a certificate:
```

administration/transport-security.md
```diff
@@ -171,3 +172,41 @@ Fluent Bit supports [TLS server name indication](https://en.wikipedia.org/wiki/S
...
+
+```
+X509v3 Subject Alternative Name:
+    DNS:my.fluent-aggregator.net
+```
+
+Then, the TLS client will connect to `other.fluent-aggregator.net`, there is a glitch of hostnames.
```
patrick-stephens, line 185:
```suggestion
As you can see, this certificate covers only `my.fluent-aggregator.net` so if we use a different hostname it should fail.
```

administration/transport-security.md
```diff
@@ -171,3 +172,41 @@ Fluent Bit supports [TLS server name indication](https://en.wikipedia.org/wiki/S
...
+
+This difference of the hostname can be detected with:
```
patrick-stephens, line 187:
```suggestion
To fully verify the alternative name and demonstrate the failure we enable `tls.verify_hostname`:
```

patrick-stephens (commented):
Is there a reason we default to not verifying the alternative name? This seems counter to an expectation of secure by default.

cosmo0920:
> Is there a reason we default to not verifying the alternative name? This seems counter to an expectation of secure by default.

Once, I implemented this feature as forcibly style, I just though such way. Because hostname verification is mostly required for TLS connections to prevent malformed hostname attached instances. However, we got realized that there is many issues in certificates for kubelet. So, preserving the backward compatibilities, we need to disable by default.

patrick-stephens:
> > Is there a reason we default to not verifying the alternative name? This seems counter to an expectation of secure by default.
> 
> Once, I implemented this feature as a forcibly style, I just though such way. Because hostname verification is mostly required for TLS connections to prevent malformed hostname attached instances. However, we got realized that there is many issues in certificates for kubelet. So, preserving the backward compatibilities, we need to disable by default.

Is this just to handle kubelet specific things then? I'd expect tls verify for everything else, e.g. outputs, etc., so can we just add a special config for kubelet usage to disable it?

cosmo0920:
> > > Is there a reason we default to not verifying the alternative name? This seems counter to an expectation of secure by default.
> > 
> > 
> > Once, I implemented this feature as a forcibly style, I just though such way. Because hostname verification is mostly required for TLS connections to prevent malformed hostname attached instances. However, we got realized that there is many issues in certificates for kubelet. So, preserving the backward compatibilities, we need to disable by default.
> 
> Is this just to handle kubelet specific things then? I'd expect tls verify for everything else, e.g. outputs, etc., so can we just add a special config for kubelet usage to disable it?

Hmm.., I see. For 3.x line, we need to restore the behavior. But it's reasonable to verify certificates as much as possible in 3.1. 
@edsiper Can we break the behavior around certificates in Fluent Bit 3.1? It's reasonable to verify certificates except for filter_kubernetes.

cosmo0920:
Oh, merged my PRs as-is. Also, users complains in in_forward and filter_kubernetes. This is because our network stack is not dumped for its error status with human readable errors. So, we need to restore the behavior. For TLS client, we can print the error message from OpenSSL's certificate verification. But, for server, we didn't get how to print an error with human readable modifications.

lecaros (approved):

PR #1395, opened 2024-06-18
main: Bump to 3 1

lecaros:


patrick-stephens (approved):
Presumably not to be merged until we have 3.1 releases going?

lecaros:
@patrick-stephens  The main branch is in _dev mode_ now, meaning we're contributing docs for the next release. The dev version is also published (https://docs.fluentbit.io/manual/v/3.1-dev) , just it's not the default one. It works the same as the main branch of Fluent Bit.

PR #1399, opened 2024-06-28
in_splunk: Add keepalive parameter document and its note

cosmo0920:
Corresponding to https://github.com/fluent/fluent-bit/pull/9019.


cosmo0920:
This PR should be invalid due to not introducing keepalive parameter for in_splunk. This parameter is unified in downstream's config_map.

PR #1400, opened 2024-06-28
Eks pod identity

iandrewt:
See https://github.com/fluent/fluent-bit/pull/9013 :)


administration/aws-credentials.md
```diff
@@ -27,12 +27,17 @@ At this time, no other settings are supported.
 Fetches credentials via a signed web identity token for a Kubernetes service account.
 See https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html.
 
-## 4. ECS HTTP Credentials Endpoint
+## 4. EKS Pod Identities (HTTP Credentials)
+
+Fetches credentials via an agent that runs on the EKS node.
+See https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html.
```
cnorris-cs, line 33: Our style would have us avoid words like "via" and to not use bare URLs.

```suggestion
Fetches credentials using an agent running on the EKS node. For more information, see
[EKS Pod Identities](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html)
in the Amazon EKS User Guide.
```

iandrewt: Would you like me to change the other linked entries to use the `For more information, see [type](link) in the Amazon <service> User Guide.` layout too? Given this page is currently not indexed, it's probably a good idea to get this page right before we add it

cnorris-cs: @iandrewt - Please. :)

PR #1411, opened 2024-07-16
Typo in secure forward key

claudex:
Fix a small typo in user key definition in Forward protocol input.


lockewritesdocs (approved):
LGTM 🦖 Thanks for the fix @claudex!

PR #1414, opened 2024-07-24
input: kafka: Remove duplicate poll_ms config

sh-cho:


cnorris-cs (approved):
Looks good from a technical writing perspective.

PR #1418, opened 2024-08-05
Add scarf pixels

lecaros:


administration/configuring-fluent-bit/classic-mode/configuration-file.md
```diff
@@ -2,6 +2,8 @@
 description: This page describes the main configuration file used by Fluent Bit
 ---
 
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=5e67142e-3887-4b56-b940-18494bcc23a7" />
```
cnorris-cs, line 5: We should be consistent in where we add the image. Since it follows the h1 in most of the modified files, it should continue to be after the h1 in all of the files.

administration/configuring-fluent-bit/classic-mode/variables.md
```diff
@@ -1,6 +1,7 @@
 # Variables
 
 Fluent Bit supports the usage of environment variables in any value associated to a key when using a configuration file.
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=1731c7b5-34c6-424f-bfc6-88c2aa71e81f" />
```
cnorris-cs, line 4: Same here. Since it was h1, blank line, scarf image, blank line, and then content, we should do that here as well.

administration/backpressure.md
```diff
@@ -1,5 +1,7 @@
 # Backpressure
 
+<img referrerpolicy="no-referrer-when-downgrade" src="https://static.scarf.sh/a.png?x-pxid=63e37cfe-9ce3-4a18-933a-76b9198958c1" />
```
cnorris-cs, line 3: It may be helpful to add HTML comments on the lines preceding these images to state what they're being used for, so that we can both be transparent, and so we can avoid people from trying to strip these from the files.

cnorris-cs (approved):
I left comments, but I've added my technical writing-based approval based on that to keep this moving.

administration/backpressure.md
lecaros: it makes sense. I'm adding it to my ToDo. Thanks for the feedback @cnorris-cs!

PR #1430, opened 2024-08-13
pipeline: inputs: kubernetes-events: Add indent in example config

sh-cho:
All entries in under sections(ex. SERVICE, INPUT, ...) must be indented, 4 spaces ideally.


alexakreizinger (approved):
good catch, thanks for fixing that!

PR #1434, opened 2024-08-15
pipeline: inputs: kubernetes-events: Fix typo

sh-cho:
prece'n'dence -> precedence

### DIFF
--- a/pipeline/inputs/kubernetes-events.md
+++ b/pipeline/inputs/kubernetes-events.md
@@ -50,4 +50,10 @@ In the following configuration file, the input plugin *kubernetes_events* collec
 ```
 
 ### Event Timestamp
-Event timestamp will be created from the first existing field in the following order of precedence: lastTimestamp, firstTimestamp, metadata.creationTimestamp
+
+Event timestamps are created from the first existing field, based on the following
+order of precedence:
+
+1. `lastTimestamp`
+1. `firstTimestamp`
+1. `metadata.creationTimestamp`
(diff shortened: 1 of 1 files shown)


pipeline/inputs/kubernetes-events.md
```diff
@@ -50,4 +50,4 @@ In the following configuration file, the input plugin *kubernetes_events* collec
 ```
 
 ### Event Timestamp
-Event timestamp will be created from the first existing field in the following order of precendence: lastTimestamp, firstTimestamp, metadata.creationTimestamp
+Event timestamp will be created from the first existing field in the following order of precedence: lastTimestamp, firstTimestamp, metadata.creationTimestamp
```
cnorris-cs, line 53: Added line wrapping, removed passive voice, and turned the list into an ordered list to better indicate the order of precedence.

```suggestion

Event timestamps are created from the first existing field, based on the following
order of precedence:

1. `lastTimestamp`
1. `firstTimestamp`
1. `metadata.creationTimestamp`
```

cnorris-cs: As a note, Markdown will turn the "1"s into "1, 2, 3".

cnorris-cs (approved):
I left a suggestion, but I've added my approval based on that to keep this moving.

pipeline/inputs/kubernetes-events.md
sh-cho: Thanks!

PR #1441, opened 2024-08-21
Reindent configuration file example, improve readability, etc.

sh-cho:
- Reindent fluent-bit (classic) configuration file example with 4 spaces
  - Entries under section must be indented, 4 spaces ideally ([docs](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/classic-mode/format-schema#sections))
  - You don't necessarily need to make it 4 spaces (any same indent level in single section is OK -just like python), but many people copy the example code from the documentation and try to use it, so IMO it's better to make it ideal.
- Add empty line between sections in config file
- Add syntax highlight for yaml (` ```yml`)
- Remove 'Line separator(U+2028)' character from vivo-exporter

### DIFF
--- a/pipeline/filters/kubernetes.md
+++ b/pipeline/filters/kubernetes.md
@@ -314,27 +314,27 @@ The difference is that kubelet need a special permission for resource `nodes/pro
 Fluent Bit Configuration Example:
 
 ```text
-    [INPUT]
-        Name              tail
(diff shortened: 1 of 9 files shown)


lockewritesdocs (approved):
LGTM 🦖 Thanks @sh-cho!

PR #1448, opened 2024-08-27
Add workers info

alexakreizinger:
this PR adds info about threading to each applicable output doc.


esmerel (approved):
wow that looks tedious.

PR #1457, opened 2024-09-12
Fluent bit: docs: Updating validating data for style

esmerel:


local-testing/validating-your-data-and-structure.md
```diff
@@ -1,55 +1,90 @@
...
-![](../.gitbook/assets/flb_pipeline_simplified_expect.png)
+tail --> grep
+grep --> record
+record --> stdout
+```
 
-Expect filter sets rules that aims to validate certain criteria like:
+Add data validation between each step to ensure your data structure is correct.
+
+This example uses the `expect` filter.
```
cnorris-cs, line 49: I really don't know on this one, and it may be out of scope of these changes. Is it Expect or expect? I've seen both capitalizations in this doc.

cnorris-cs (approved):
I left a comment, but I've added my approval based on that to keep this moving.

local-testing/validating-your-data-and-structure.md
esmerel: I tried to use caps at the beginning of sentences, but lowercase everywhere else. because my brain objected to lowercase at the beginning of sentences.

PR #1459, opened 2024-09-12
Remove "Visualize" info

alexakreizinger:
this PR removes information about using the Calyptia UI to visualize Fluent Bit configuration files, which used to be an option available to free/trial users but no longer exists.


cnorris-cs (approved):
Looks good. Keeping the scope of my comments to literally the removal of the visualize info.

PR #1470, opened 2024-10-10
Fix `ddtags` missing backtick in Datadog

edgarrmondragon:


esmerel (approved):
Thanks for catching that!

esmerel (approved):
Thanks for catching that!

PR #1474, opened 2024-10-17
docs: testing: add vale styles

esmerel:
This PR adds Vale, an open source docs linter. This is aimed at helping ensure spell checks, consistent wording, and a number of general grammar checks based on Google's style guide.

These tests are all at the 'suggestion' level, and will are not required. They should not break any testing, and vale is limited to running on the following file types: md,markdown,txt,htm,html.


cnorris-cs (approved):
Several of the lines at the end of the exception list aren't needed, but they also don't hurt anything. I'm approving this so that Vale tests of the content can happen.

PR #1477, opened 2024-10-18
docs: pipeline: Updating record-modifier for style

esmerel:
Updates the record-modifier plugin page for consistent grammar and style.


cnorris-cs (approved):
I left a suggestion, but I've added my approval based on that to keep this moving.

PR #1478, opened 2024-10-21
pipelines: docs: bringing elasticsearch up to standards

esmerel:


lockewritesdocs (approved):
I pushed some minor changes like adding _none_ to the Default column where no value exists instead of suggesting it in the PR 10 times.

PR #1488, opened 2024-10-24
concepts: parsers: update page for style

esmerel:
Applying style for consistency.

### DIFF
--- a/concepts/data-pipeline/parser.md
+++ b/concepts/data-pipeline/parser.md
@@ -29,7 +29,7 @@ following Apache (HTTP Server) log entry:
 192.168.2.20 - - [28/Jul/2006:10:27:10 -0300] "GET /cgi-bin/try/ HTTP/1.0" 200 3395
 ```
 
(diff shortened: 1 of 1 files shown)


concepts/data-pipeline/parser.md
```diff
@@ -19,13 +21,18 @@ graph LR
...
-```
+The parser converts unstructured data to structured data. As an example, consider the
+following Apache (HTTP Server) log entry:
+
+```text
 192.168.2.20 - - [28/Jul/2006:10:27:10 -0300] "GET /cgi-bin/try/ HTTP/1.0" 200 3395
 ```
 
-The above log line is a raw string without format, ideally we would like to give it a structure that can be processed later easily. If the proper configuration is used, the log entry could be converted to:
+This log line is a raw string without format. Giving it a structure makes it easier
```
alexakreizinger, line 32:
```suggestion
This log line is a raw string without format. Structuring the log makes it easier
```

alexakreizinger (approved):

PR #1501, opened 2024-11-05
admin: proxy: update http proxy for style

esmerel:


administration/http-proxy.md
```diff
@@ -1,63 +1,81 @@
...
 When no authentication is required, omit the username and password:
 
-```bash
+```text
 HTTP_PROXY='http://proxy.example.com:8080'
 ```
 
-The `HTTP_PROXY` environment variable is a [standard way](https://docs.docker.com/network/proxy/#use-environment-variables) for setting a HTTP proxy in a containerized environment, and it is also natively supported by any application written in Go. Therefore, we follow and implement the same convention for Fluent Bit. For convenience and compatibility, the `http_proxy` environment variable is also supported. When both the `HTTP_PROXY` and `http_proxy` environment variables are provided, `HTTP_PROXY` will be preferred.
+The `HTTP_PROXY` environment variable is a [standard
+way](https://docs.docker.com/network/proxy/#use-environment-variables) for setting a
```
cnorris-cs, line 30:
```suggestion
method](https://docs.docker.com/network/proxy/#use-environment-variables) for setting a
```

administration/http-proxy.md
```diff
@@ -1,63 +1,81 @@
...
+proxy. The `no_proxy` environment variable is also supported. When both `NO_PROXY`
+and `no_proxy` environment variables are provided, `NO_PROXY` takes precedence.
 
-The format for the no proxy environment variable is a comma-separated list of hostnames or IP addresses whose traffic should not flow through the HTTP proxy.
+The format for the `no_proxy` environment variable is a comma-separated list of
+host names or IP addresses.
 
-A domain name matches itself and all its subdomains (i.e. `foo.com` matches `foo.com` and `bar.foo.com`):
+A domain name matches itself and all its subdomains (for example:`example.com` matches
+`example.com` and `test.example.com`):
```
cnorris-cs, line 58:
```suggestion
A domain name matches itself and all of its subdomains (for example,`example.com`
matches both `example.com` and `test.example.com`):
```

administration/http-proxy.md
```diff
@@ -1,63 +1,81 @@
...
 
-```bash
+```text
 NO_PROXY='foo.com,127.0.0.1,localhost'
 ```
 
-A domain with a leading `.` only matches its subdomains (i.e. `.foo.com` matches `bar.foo.com` but not `foo.com`):
+A domain with a leading dot (`.`) only matches its subdomains (for example:
```
cnorris-cs, line 64:
```suggestion
A domain with a leading dot (`.`) matches only its subdomains (for example:
```

cnorris-cs (approved):

PR #1526, opened 2024-12-04
Adding GCS info to S3 output plus bare basics on style

esmerel:
This PR adds the info from https://github.com/fluent/fluent-bit/issues/4165 about how to send S3 data to Google.

Also did a super basic cleanup of the easiest to fix markdown and vale problems.


lockewritesdocs (approved):
LGTM 🦖

PR #1576, opened 2025-03-01
[kubernetes_filter]: Update documentation with the usage of Owner_References.

shuaich:
Querying owner references via kubernetes filter is merged in https://github.com/fluent/fluent-bit/pull/9719.


shuaich:
Have no idea why docker.md and windows.md are also modified. I only updated kuberentes filter.

PR #1585, opened 2025-03-10
stackdriver: Refactor special fields docs

braydonk:
The special fields documentation as it existed was a bit of a mishmash of old documentation for a previous logging agent solution and a lot of additions over the years trying to match the old format. It ends up reproducing a lot of information that is better stated by the existing Cloud Logging docs, while missing crucial details about how the plugin handled these cases. Certain fields were also missing.

This PR rewrites the documentation, doing more linking out to Google Cloud Logging documentation in place of rewording the same information. It also guides more generally on some of the edge cases for how the special fields are handled, calling out the exceptions in a clearer way. It also adds some fields that were previously missing from this document.

### DIFF
--- a/pipeline/outputs/stackdriver.md
+++ b/pipeline/outputs/stackdriver.md
@@ -55,7 +55,7 @@ If you are using a _Google Cloud Credentials File_, the following configuration 
 
 Example configuration file for k8s resource type:
 
-local_resource_id is used by stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the local_resource_id field in the log entry. If there is no field logging.googleapis.com/local_resource_id in the log, the plugin will then construct it by using the tag value of the log.
+`local_resource_id` is used by stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the local_resource_id field in the log entry. If there is no field logging.googleapis.com/local_resource_id in the log, the plugin will then construct it by using the tag value of the log.
 
 The local_resource_id should be in format:
 
--- a/pipeline/outputs/stackdriver.md
+++ b/pipeline/outputs/stackdriver.md
@@ -157,6 +157,15 @@ For instance, for a K8s resource type, `resource_labels` can be used in tandem w
 ```
 
 `resource_labels` also supports validation for required labels based on the input resource type. This allows fluent-bit to check if all specified labels are present for a given configuration before runtime. If validation is not currently supported for a resource type that you would like to use this API with, we encourage you to open a pull request for it. Adding validation for a new resource type is simple - all that is needed is to specify the resources associated with the type alongside the required labels [here](https://github.com/fluent/fluent-bit/blob/master/plugins/out_stackdriver/stackdriver_resource_types.c#L27).
+
+## Log Name
+
+By default, the plugin will write to the following log name:
+```
+/projects/<project ID>/logs/<log tag>
+```
+You may be in a scenario where being more specific about the log name is important (for example [integration with Log Router rules](https://cloud.google.com/logging/docs/routing/overview) or [controlling cardinality of log based metrics]((https://cloud.google.com/logging/docs/logs-based-metrics/troubleshooting#too-many-time-series))). You can control the log name directly on a per-log basis by using the [`logging.googleapis.com/logName` special field][StackdriverSpecialFields]. You can configure a `log_name_key` if you'd like to use something different than `logging.googleapis.com/logName`, i.e. if the `log_name_key` is set to `mylognamefield` will extract the log name from `mylognamefield` in the log.
+
 ## Troubleshooting Notes
 
 ### Upstream connection error
--- a/pipeline/outputs/stackdriver.md
+++ b/pipeline/outputs/stackdriver.md
@@ -55,7 +55,7 @@ If you are using a _Google Cloud Credentials File_, the following configuration 
 
 Example configuration file for k8s resource type:
 
-`local_resource_id` is used by stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the local_resource_id field in the log entry. If there is no field logging.googleapis.com/local_resource_id in the log, the plugin will then construct it by using the tag value of the log.
+`local_resource_id` is used by the Stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the `local_resource_id` field in the log entry. If there is no field `logging.googleapis.com/local_resource_id` in the log, the plugin will then construct it by using the tag value of the log.
 
 The local_resource_id should be in format:
 
--- a/pipeline/outputs/stackdriver_special_fields.md
+++ b/pipeline/outputs/stackdriver_special_fields.md
@@ -221,7 +221,7 @@ The `logging.googleapis.com/monitored_resource` field is parsed in a special way
 
 The `type` field from the [`MonitoredResource` proto]() is not parsed out of the special field. It is read from the [`resource` plugin configuration option](https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver#configuration-parameters). If it is supplied in the `logging.googleapis.com/monitored_resource` special field, it will not be recognized.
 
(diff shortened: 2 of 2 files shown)


pipeline/outputs/stackdriver_special_fields.md
```diff
@@ -5,187 +5,142 @@ When the [google-logging-agent](https://cloud.google.com/logging/docs/agent) rec
...
-| logging.googleapis.com/sourceLocation | sourceLocation | Additional information about the source code location that produced the log entry. |
-| logging.googleapis.com/http_request | httpRequest | A common proto for logging HTTP requests. |
-| logging.googleapis.com/trace | trace | Resource name of the trace associated with the log entry |
-| logging.googleapis.com/traceSampled | traceSampled | The sampling decision associated with this log entry. |
-| logging.googleapis.com/spanId | spanId | The ID of the trace span associated with this log entry. |
-| timestamp | timestamp | An object including the seconds and nanos fields that represents the time |
-| timestampSecond & timestampNanos | timestamp | The seconds and nanos that represents the time |
+| JSON log field | [LogEntry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) field | type | Description |
+| :--- | :--- | :--- | :--- |
+| `logging.googleapis.com/logName` | `logName` | `string` | The log name to write this log to. |
```
ensonic, line 10: Should we mention here or in this doc that each of those fields must no exceed a cardinality of 30k (https://cloud.google.com/logging/docs/logs-based-metrics/troubleshooting#too-many-time-series) if one wants to use log based metrics?

pipeline/outputs/stackdriver.md
```diff
@@ -55,7 +55,7 @@ If you are using a _Google Cloud Credentials File_, the following configuration
 
 Example configuration file for k8s resource type:
 
-local_resource_id is used by stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the local_resource_id field in the log entry. If there is no field logging.googleapis.com/local_resource_id in the log, the plugin will then construct it by using the tag value of the log.
+`local_resource_id` is used by stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the local_resource_id field in the log entry. If there is no field logging.googleapis.com/local_resource_id in the log, the plugin will then construct it by using the tag value of the log.
```
esmerel, line 58:
```suggestion
`local_resource_id` is used by the Stackdriver output plugin to set the labels field for different k8s resource types. Stackdriver plugin will try to find the `local_resource_id` field in the log entry. If there is no field `logging.googleapis.com/local_resource_id` in the log, the plugin will then construct it by using the tag value of the log.
```

pipeline/outputs/stackdriver_special_fields.md
```diff
@@ -195,104 +150,44 @@ the logEntry will be:
...
     ...
 }
 ```
-The logging agent will log an error and reject the entry.
-
-## SourceLocation
-SourceLocation field contains additional information about the source code location that produced the log entry. The format.
 
-The JSON representation is as followed:
+**If the field itself is not a map, the plugin will leave this field untouched.** For example:
```
esmerel, line 161:
```suggestion
If the field itself is not a map, the plugin will leave this field untouched. For example:
```

pipeline/outputs/stackdriver_special_fields.md
```diff
@@ -195,104 +150,44 @@ the logEntry will be:
...
-    }
     ...
 }
 ```
 
-## httpRequest
-HttpRequest field is a common proto for logging HTTP requests.
-
-The JSON representation is as followed:
+**If there are extra subfields, the plugin will add the recognized fields to the corresponding field in the LogEntry, and preserve the extra subfields in jsonPayload.** For example:
```
esmerel, line 179:
```suggestion
If there are extra subfields, the plugin will add the recognized fields to the corresponding field in the LogEntry, and preserve the extra subfields in jsonPayload. For example:
```

pipeline/outputs/stackdriver_special_fields.md
```diff
@@ -301,75 +196,41 @@ the logEntry will be:
...
-```text
-"trace": string
-```
+The `logging.googleapis.com/monitored_resource` field is parsed in a special way, meaning it has some important exceptions:
 
-## Span ID
+The `type` field from the [`MonitoredResource` proto]() is not parsed out of the special field. It is read from the [`resource` plugin configuration option](https://docs.fluentbit.io/manual/pipeline/outputs/stackdriver#configuration-parameters). If it is supplied in the `logging.googleapis.com/monitored_resource` special field, it will not be recognized.
 
-The span ID within the trace associated with the log entry.
+The `labels` field is expected to be an `object<string, string>`. If any fields have a value that is not a string, the value will be ignored and **not preserved**. The plugin will log an error and drop the field.
```
esmerel, line 224:
```suggestion
The `labels` field is expected to be an `object<string, string>`. If any fields have a value that is not a string, the value will be ignored and not preserved. The plugin logs an error and drops the field.
```

esmerel (commented):
lots of good updates here! I'm only not approving it because of the one spot where it looks like a URL is missing and I don't want to see broken render or bad links =)

pipeline/outputs/stackdriver_special_fields.md
braydonk: The particular way that log name happens to interact with log based metrics strikes me as something Fluent Bit itself shouldn't be concerned about, so I wasn't sure the right way to bring that in. I think I found a nice way to bring it in, while still mostly leaning on linking out to our own docs on the subjects. In [40631d3](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/40631d3fc506102f79ef4b5a4364c5a47d7b75a9) I add a new section to the main stackdriver plugin docs about default Log Name behaviour, along with how to customize it and some examples for why you may want to do this. Hopefully this should get across the point for anyone else who may encounter the same problem.

Side note: I think our own docs on this aren't amazing on this front either, and could probably do with some polishing. I'll raise it with the technical writers who (I think might) work on this internally.

braydonk:
> the one spot where it looks like a URL is missing

Oops! Thanks, fixed in [40631d3](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/40631d3fc506102f79ef4b5a4364c5a47d7b75a9).

pipeline/outputs/stackdriver_special_fields.md
braydonk: Fixed in [40631d3](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/40631d3fc506102f79ef4b5a4364c5a47d7b75a9)

pipeline/outputs/stackdriver_special_fields.md
braydonk: Fixed in [40631d3](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/40631d3fc506102f79ef4b5a4364c5a47d7b75a9)

braydonk:
Ah oops, I don't have automatic commit signing set up in my local environment for this repo. Force-push coming to fix the signing sorry!

pipeline/outputs/stackdriver_special_fields.md
braydonk: Fixed in [f2f67cf](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/f2f67cf3cd9d5af702c78466174accaa9e401754)

pipeline/outputs/stackdriver.md
braydonk: Fixed in [f2f67cf](https://github.com/fluent/fluent-bit-docs/pull/1585/commits/f2f67cf3cd9d5af702c78466174accaa9e401754)

There are likely a number of little cleanups to do here that I will try and find time to do. Nobody who has worked on these docs from the Google side is a technical writer by trade (nor am I) so a lot of it could be cleaner. Fixing these things ends up being sadly a bit of a side priority, but I will still try and come back and clean stuff up when the opportunity presents itself.

braydonk:
All open feedback should be addressed now, thanks!

pipeline/outputs/stackdriver_special_fields.md
ensonic: Thanks a lot! An extra section makes a lot of sense.

patrick-stephens (approved):

PR #1598, opened 2025-03-30
conditional: add conditional processing documentation

niedbalski:
This PR adds documentation for the conditional processing feature for processors, which allows selective application of processors based on field values. It includes configuration details, available operators, and examples.

### DIFF
--- a/pipeline/processors/conditional-processing.md
+++ b/pipeline/processors/conditional-processing.md
@@ -1,6 +1,8 @@
 # Conditional processing
 
-Conditional processing allows you to selectively apply processors to log records based on field values. This feature enables you to create processing pipelines that apply processors only to records that match specific criteria.
+Conditional processing lets you selectively apply [processors](README.md) to logs based on
+the value of fields that those logs contain. This feature lets you create processing pipelines
+that only process records that meet certain criteria, and ignore the rest.
 
 ## Configuration
 


pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
 # Processors
 
+Processors are components that can modify, transform, or enhance data records as they flow through the Fluent Bit pipeline.
+
+## Available processors
+
+Fluent Bit offers the following processors:
```
patrick-stephens, line 7: We should say that all existing legacy filters can be processors too.

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
 # Processors
 
+Processors are components that can modify, transform, or enhance data records as they flow through the Fluent Bit pipeline.
```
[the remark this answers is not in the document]
leonardo-albertovich: Definitely, let me try to give you a brief explanation starting with 2 important things : 

1. Filters can act as processors
2. Processors **cannot** act as filters

This limitation stems from the design choices and the fact that while processors are the new and preferred way to operate we understood that it didn't make sense to lose access to the existing filter corpus.

In both of this cases the pipeline would look something like this : 

.Filters : input -> filter & persistence -> output
.Processors : input -> processor stack -> filter & persistence -> output

The key aspect here is that processors are tightly coupled to inputs which means they are executed immediately when ingesting data which allows us to offload the CPU time required to execute those tasks to the input plugins thread rather than creating a centralized bottleneck.

This is particularly important for regular expressions as those are really heavy and used to cause significant timeslice shifting.

One of the weak spots of processors (before conditional processing was added I think) was that anything that passed through the processor stack was acted on (in contrast to how filters actually match based on the tag) but that's a thing of the past.

There is an additional use case for filters that are directly attached to output plugins but I don't think I can give any meaningful examples at the moment.

Feel free to ask for clarifications or expansion on anything you need in order to enrich this section.

Thanks.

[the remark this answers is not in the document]
alexakreizinger: that's great info, thank you :) 

@niedbalski I'm not sure what our timeline is for releasing this feature/associated docs, but if we can wait at least a few days (since that's how long I'll likely need), I'm happy to take a thorough look at this PR and make any additions or edits for clarity, etc. since I wasn't involved in the development of this at all, I'm looking at it from the perspective of a brand-new user 😆

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+
+## Available processors
+
+Fluent Bit offers the following processors:
+
+- [Content Modifier](content-modifier.md): Manipulate message content, metadata/attributes for logs and traces
```
alexakreizinger, line 9:
```suggestion
- [Content Modifier](content-modifier.md): Manipulate the content, metadata, and attributes of logs and traces.
```

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+- [Labels](labels.md): Add, update or delete labels in records
```
alexakreizinger, line 10:
```suggestion
- [Labels](labels.md): Add, update, or delete metric labels.
```

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+- [Metrics Selector](metrics-selector.md): Select specific metrics
```
alexakreizinger, line 11:
```suggestion
- [Metrics Selector](metrics-selector.md): Choose which metrics to keep or discard.
```

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+- [OpenTelemetry Envelope](opentelemetry-envelope.md): Convert logs to OpenTelemetry format
```
alexakreizinger, line 12:
```suggestion
- [OpenTelemetry Envelope](opentelemetry-envelope.md): Transform logs into an OpenTelemetry-compatible format.
```

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+- [SQL](sql.md): Process records using SQL queries
```
alexakreizinger, line 13:
```suggestion
- [SQL](sql.md): Use SQL queries to extract log content.
```

pipeline/processors/conditional-processing.md
```diff
@@ -0,0 +1,207 @@
+# Conditional processing
+
+Conditional processing allows you to selectively apply processors to log records based on field values. This feature enables you to create processing pipelines that apply processors only to records that match specific criteria.
```
alexakreizinger, line 3: does conditional processing ONLY apply to logs? (I see that certain processors apply to metrics/traces as well; not sure if that means that certain types of processors can't be used conditionally since they don't work with logs)

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+
+## Features
+
```
alexakreizinger, line 16:
```suggestion
## Features

All available processors include the following features:

```

not sure if "all available" is the correct phrasing here (since I'm [not sure if conditional processing only applies to log-centric processors](https://github.com/fluent/fluent-bit-docs/pull/1598/files#r2023827864)), but putting this as a placeholder for now.

pipeline/processors/README.md
```diff
@@ -1,2 +1,17 @@
...
+- [Conditional Processing](conditional-processing.md): Apply processors only to records that meet specific conditions
```
alexakreizinger, line 17:
```suggestion
- [Conditional Processing](conditional-processing.md): Apply processors selectively to records that meet specific criteria.
```

pipeline/processors/conditional-processing.md
```diff
@@ -0,0 +1,207 @@
...
+Conditional processing is available for processors in the YAML configuration format. To apply a processor conditionally, you add a `condition` block to the processor configuration:
+
+```yaml
+- name: processor_name
+  # Regular processor configuration...
+  condition:
+    op: and|or
+    rules:
+      - field: "$field_name"
+        op: comparison_operator
```
alexakreizinger, line 16: is it too late to change how this feature is designed so that `condition.op` and `condition.rules.op` have different names? trying to distinguish between the two/explain how they're different is a bit difficult.

(maybe `condition.op` would be `global_operator` and `condition.rules.op` would be `rule_operator`...)

alexakreizinger:
image
not sure why this didn't pass the DCO check, since it seems to be signed off correctly.

(necessary commit to make the page visible in the sidebar, though. also, without it, the link to this page in `pipeline/processors/README.md` breaks.)

pipeline/processors/conditional-processing.md
```diff
@@ -0,0 +1,207 @@
...
+| `not_regex` | Does not match regular expression |
+| `in` | Value is in the specified array |
+| `not_in` | Value is not in the specified array |
+
+### Field access
+
+You can access record fields using [record accessor syntax](/administration/configuring-fluent-bit/classic-mode/record-accessor.md):
+
+- Simple fields: `$field`
+- Nested fields: `$parent['child']['subchild']`
```
alexakreizinger, line 60:
```suggestion
## Configuration

You can turn a standard processor into a conditional processor by adding a `condition` block to the
processor's YAML configuration settings.

{% hint style="info" %}
Conditional processing is only available for
[YAML configuration files](../../administration/configuring-fluent-bit/yaml/README.md),
not [classic configuration files](../../administration/configuring-fluent-bit/classic-mode/README.md).
{% endhint %}


These `condition` blocks use the following syntax:

```yaml
pipeline:
  inputs:
  <...>
      processors:
        logs:
          - name: {processor_name}
            <...>
            condition:
	      op: {and|or}
              rules:
                - field: {field_name1}
                  op: {comparison_operator}
                  value: {comparison_value1}
                - field: {field_name2}
                  op: {comparison_operator}
                  value: {comparison_value2}
            <...>
```

Each processor can only have a single `condition` block, but can have multiple rules within that condition.
These rules are stored as items in the `condition.rules` array.

### Condition evaluation

The `condition.op` parameter specifies the condition's evaluation logic. It has two possible values:

- `and`: All rules in the `condition.rules` array must evaluate to `true` for the condition to be met.
- `or`: One or more rules in the `conditions.rules` array must evaluate to `true` for the condition to be met.

### Rule parameters

Each item in the `condition.rules` array must include values for the following parameters:

| Parameter | Description |
| --- | --- |
| `field` | The field within your logs to evaluate. The value of this parameter must use [the correct syntax](#field-access) to access the fields inside logs. |
| `op` | The [comparison operator](#comparison-operators) to evaluate whether the rule is true. This parameter (`condition.rules.op`) is distinct from the `condition.op` parameter and has different possible values. |
| `value` | The value of the specified log field to use in your comparison. | 

#### Field access

The `conditions.rules.field` parameter uses [record accessor syntax](/administration/configuring-fluent-bit/classic-mode/record-accessor.md) to reference fields inside logs.

You can use `$field` syntax to access a top-level field, and `$field['child']['subchild']` to access nested fields.

#### Comparison operators

The `conditions.rules.op` parameter has the following possible values:

- `eq`: equal to
- `neq`: not equal to
- `gt`: greater than
- `lt`: less than
- `gte`: greater than or equal to
- `lte`: less than or equal to
- `regex`: matches a regular expression
- `not_regex`: does not match a regular expression
- `in`: is included in the specified array
- `not_in`: is not included in the specified array
```

File: pipeline/processors/conditional-processing.md [hunk 13]
```diff
@@ -0,0 +1,207 @@
...
+
+## Examples
+
+### Simple condition
+
+Process records only when the HTTP method is POST:
```
alexakreizinger, line 66:
```suggestion
### Basic condition

This example applies a condition that only processes records with the HTTP method `POST`:
```

pipeline/processors/conditional-processing.md
```diff
@@ -0,0 +1,207 @@
...
+                  op: gt
+                  value: 100
+                - field: "$request['status_code']"
+                  op: gte
+                  value: 400
+```
+
+### Using IN operator
+
+Apply a processor when a value matches one of multiple options:
```
alexakreizinger, line 144:
```suggestion
### Array of values

This example uses an array for the value of `condition.rules.value`:
```

pipeline/processors/conditional-processing.md
```diff
@@ -0,0 +1,207 @@
...
+              op: and
+              rules:
+                - field: "$request['method']"
+                  op: in
+                  value: ["POST", "PUT", "DELETE"]
+```
+
+## Multiple processors with conditions
+
+You can chain multiple conditional processors to create advanced processing pipelines:
```
alexakreizinger, line 168:
```suggestion
### Multiple processors with conditions

This example uses multiple processors with conditional processing enabled for each:
```

pipeline/processors/conditional-processing.md
alexakreizinger, line 3: conditional (pun not intended) suggestion for now, but if it turns out that conditional processing also applies to metrics/traces, I'll need to edit this accordingly (and a few other places throughout the doc).
```suggestion
Conditional processing lets you selectively apply [processors](README.md) by determining whether
to process logs based on the value of fields they contain. This feature lets you create processing pipelines
that only process records that meet certain criteria.
```

PR #1603, opened 2025-04-19
installation: windows: Fix Service creation C:\Program Files via PowerShell

sier:
As `C:\Program Files\fluent-bit\` is the default install location for fluent-bit it makes sense that the documentation supports this.
The current work-around doesn't make sense when it can easily be resolved by simply escaping the quotation marks.
This should give new users a better experience as the documentation follows a default installation.

Feel free to remove/ignore the -Description commit, I just think it gives a more concise and proper look inline with how most Windows services look and feel.


esmerel (approved):
Seems completely reasonable to me.

PR #1612, opened 2025-04-24
stream-processing: overview: basic cleanup

alexakreizinger:


alexakreizinger:
accidentally made changes for a different doc in the same PR, but since they're closely related I'll just leave it :P

PR #1619, opened 2025-04-29
Change to name Sampling instead of Traces in documentation. Fixes #1618.

### LINKED ISSUE #1618: Rename the tracing page to sampling
Eduardo requesting update to conform the tracing docs to use the same 'name' convention, so this needs to be updated to 'sampling'. 

Please assign to me!

eschabell:
Moved traces to sampling naming convention in docs.


patrick-stephens (commented):
Change itself looks fine but any reason we changed all the bulleted list types?

edsiper:
I recommend not touching files that are not part of the main change

esmerel:
> Change itself looks fine but any reason we changed all the bulleted list types?

I'm not sure it's necessary as part of this change but we recommend using - instead of * for bulleted lists in general. It prevents confusion and accidents with italics or bolding.

eschabell:
There is a single change in that file, but it looks like my code editor auto "fixed" that.... must be some standard. I'll fix that and update the PR.

patrick-stephens (approved):

PR #1626, opened 2025-05-01
stream-processing: README: general cleanup

alexakreizinger:


esmerel (approved):
one question but otherwise lgtm

PR #1628, opened 2025-05-02
stream-processing: getting-started: check-keys-null-values: general cleanup

alexakreizinger:


stream-processing/getting-started/check-keys-null-values.md
```diff
@@ -1,38 +1,34 @@
-# Check Keys and NULL values
+# Check keys and null values
 
-> Feature available on Fluent Bit &gt;= 1.2
+When working with structured messages (also known as records), there are certain cases where you might want to confirm whether a key exists, and whether its value is null or not null.
 
-When working with structured messages \(records\), there are certain cases where we want to know if a key exists, if it value is _null_ or have a value different than _null_.
+In Fluent Bit, records are a binary serialization of maps with keys and value. A value can be null, which is a valid data type. The following statements can be applied in Fluent Bit SQL:
```
esmerel, line 5:
```suggestion
In Fluent Bit, records are a binary serialization of maps with keys and value. A value can be `null`, which is a valid data type. The following statements can be applied in Fluent Bit SQL:
```

Not sure if this should actually be code tagged here or not. I think -I- would but I don't know that is has to be.

esmerel (approved):

PR #1640, opened 2025-05-09
Vale: Set LyHyphens to error

esmerel:
Updates the LyHyphens test to error level, fixes the single instance of error, and adds the test info to the contribution file.


cnorris-cs (approved):
Looks good.

PR #1641, opened 2025-05-09
Vale: Set Ordinal test to error, fix problems

esmerel:
This changes the Ordinal vale test to error, corrects the files with issues, and updates the contributing file with the test info.


cnorris-cs (approved):
Looks good.

PR #1642, opened 2025-05-09
Vale: Add HeadingPunctuation test to error

esmerel:
Sets the vale HeadingPunctuation test to error, corrects the one instance where this was a problem, and adds the test info to the contributions file.


cnorris-cs (approved):
Looks good.

PR #1652, opened 2025-05-14
admin/config/yaml/pipeline: Vale: Updates for style

esmerel:
Updates the selected page to meet basic style guidelines.


administration/configuring-fluent-bit/yaml/pipeline-section.md
```diff
@@ -110,9 +110,9 @@ pipeline:
               end
 ```
 
-You might noticed that processors not only can be attached to input, but also to an output.
+Processors can be attached to inputs and outputs.
 
-### How Are Processors Different from Filters?
+### How Processors are different from Filters
```
cnorris-cs, line 115: Should Processors and Filters be capitalized here? I don't think they should, since they're not capitalized in the following line in this section.

esmerel: I decided figuring out if they were proper nouns was not what this cleanup was covering right now. I think that needs a bigger picture review.

PR #1667, opened 2025-05-19
administration: add storage.backlog.flush_on_shutfown option

singholt:
### Summary

This PR douments the new storage config option `storage.backlog.flush_on_shutfown`. When enabled, Fluent Bit will attempt to flush all backlog filesystem chunks to their destination(s) during the shutdown process. This can help ensure data delivery before fluent bit stops, but may increase shutdown time.

Default value is off.

See corresponding fluent-bit PR: https://github.com/fluent/fluent-bit/pull/9952


esmerel (approved):
Looks good to me from a tech writer standpoint.

PR #1672, opened 2025-05-19
Dev: golang: Style updates

esmerel:
Update seleceted page for style.

### DIFF
--- a/development/golang-output-plugins.md
+++ b/development/golang-output-plugins.md
@@ -1,18 +1,18 @@
-# Golang Output Plugins
+# Golang output plugins
 
-Fluent Bit currently supports integration of Golang plugins built as shared objects for output plugins only. The interface for the Golang plugins is currently under development but is functional.
+Fluent Bit supports integration of Golang plugins built as shared objects for output plugins only. The interface for the Golang plugins is currently under development but is functional.
 
-## Getting Started
+## Get started
 
-Compile Fluent Bit with Golang support, e.g:
+Compile Fluent Bit with Golang support:
 
-```text
-$ cd build/
-$ cmake -DFLB_DEBUG=On -DFLB_PROXY_GO=On ../
-$ make
+```shell
+cd build/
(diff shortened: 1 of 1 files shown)


development/golang-output-plugins.md
```diff
@@ -70,17 +70,19 @@ func main() {
...
 
 ```bash
-$ go build -buildmode=c-shared -o out_gstdout.so out_gstdout.go
+go build -buildmode=c-shared -o out_gstdout.so out_gstdout.go
 ```
 
-Once built, a shared library called `out\_gstdout.so` will be available. It's really important to double check the final .so file is what we expect. Doing a `ldd` over the library we should see something similar to this:
+Once built, a shared library called `out_gstdout.so` will be available. Confirm the
+final `.so` file is as expected. When you use the `ldd` over the library  should see
+something similar to this:
```
cnorris-cs, line 83: This got accidentally wrapped.

development/golang-output-plugins.md
```diff
@@ -70,17 +70,19 @@ func main() {
...
+To build the code, use the following line:
 
 ```bash
-$ go build -buildmode=c-shared -o out_gstdout.so out_gstdout.go
+go build -buildmode=c-shared -o out_gstdout.so out_gstdout.go
 ```
 
-Once built, a shared library called `out\_gstdout.so` will be available. It's really important to double check the final .so file is what we expect. Doing a `ldd` over the library we should see something similar to this:
+Once built, a shared library called `out_gstdout.so` will be available. Confirm the
+final `.so` file is as expected. When you use the `ldd` over the library  should see
```
cnorris-cs, line 82:
```suggestion
final `.so` file is as expected. When you use the `ldd` over the library should see
```

cnorris-cs: Not that this should be wrapped, but I wanted to emphasize the double space in the line.

cnorris-cs (approved):
I left comments and suggestions, but I've added my approval based on those to keep this moving.

PR #1676, opened 2025-05-20
pipeline: inputs: tail: add inotify info

alexakreizinger:
this PR adds info about using inotify to monitor a large number of files, which I'm told has been a supported feature for a while but was never added to the docs.


gguillotte (approved):
A few non-blocking suggestions, particularly a fix of a typo predating this PR.

PR #1698, opened 2025-05-29
Pipeline: filters: standard output: style

esmerel:


pipeline/filters/standard-output.md
```diff
@@ -1,16 +1,18 @@
-# Standard Output
+# Standard output
 
-The stdout filter plugin allows printing to the standard output the data flowed through the filter plugin, which can be very useful while debugging.
+The stdout filter plugin allows printing to the standard output the data flowed through the filter plugin, which can be used while debugging.
```
alexakreizinger, line 3:
```suggestion
The standard output filter allows printing to stdout the data flowed through the filter plugin, which can be used while debugging.
```
minor thing but I think "standard output" should be the name of the filter and "stdout" the name of the actual system stream

esmerel: that's 100% legit.

PR #1701, opened 2025-05-29
Pipeline: filters: throttle: style:

esmerel:
Stylistic and consistency updates.

### DIFF
--- a/pipeline/filters/throttle.md
+++ b/pipeline/filters/throttle.md
@@ -1,6 +1,6 @@
 # Throttle
 
-The _Throttle_ filter plugin sets the average `Rate` of messages per `Interval`, based on the leaky bucket and sliding window algorithm. In case of flooding, it will leak at a certain rate.
+The _Throttle_ filter sets the average `Rate` of messages per `Interval`, based on the leaky bucket and sliding window algorithm. In case of flooding, it will leak at a certain rate.
 
 ## Configuration parameters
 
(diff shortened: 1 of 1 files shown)


pipeline/filters/throttle.md
```diff
@@ -1,17 +1,17 @@
 # Throttle
 
-The _Throttle Filter_ plugin sets the average _Rate_ of messages per _Interval_, based on leaky bucket and sliding window algorithm. In case of overflood, it will leak within certain rate.
+The _Throttle_ filter plugin sets the average `Rate` of messages per `Interval`, based on the leaky bucket and sliding window algorithm. In case of flooding, it will leak at a certain rate.
```
alexakreizinger, line 3:
```suggestion
The _Throttle_ plugin sets the average `Rate` of messages per `Interval`, based on the leaky bucket and sliding window algorithm. In case of flooding, it will leak at a certain rate.
```

pipeline/filters/throttle.md
```diff
@@ -23,7 +23,7 @@ Window 5
 Interval 1s
 ```
 
-You would received 1 message in the first second, 3 messages second, and 5 third.
+You would receive 1 message in the first second, 3 messages second, and 5 third.
 Disregard that Window is actually 5, because the configuration uses `slow` start
 to prevent flooding during the startup.
```
alexakreizinger, line 28:
```suggestion
You would receive 1 message in the first second, 3 messages second, and 5 third. Disregard that Window is actually 5, because the configuration uses `slow` start to prevent flooding during the startup.
```

pipeline/filters/throttle.md
esmerel: changing to filter for consistency.

PR #1705, opened 2025-05-30
Updating markdownlint test file format

esmerel:
Testing updating the markdownlint test file format.

Renames the old markdownlint.rb file - I don't want to remove it in case something relies on it and I'll be able to fix it easily. Updating this file wasn't actually changing test results, which tells me it wasn't being used.

Adds .markdownlint.json - Test results DO make a difference when using this file. The previous tests are represented here except for:

MD013: line length - right now, gitbook incorrectly thinks that line wraps should always be hard, and results in badly rendered pages. I'm excluding this test for now to reduce noise.

### DIFF
--- a/test-markdownlint.rb
+++ b/test-markdownlint.rb
@@ -4,12 +4,8 @@
 # markdownlint. See .markdownlint.json.
 
 # Enable all rules by default
-# all
+all
 
-## Exclude line length test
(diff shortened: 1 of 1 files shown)


cnorris-cs (approved):
Looks good.

PR #1755, opened 2025-06-18
Pipeline: input: nginx: style

esmerel:
Updating selected page for style and consistency. Removed copied metrics, and referred to the source. Copying these metrics wholesale isn't necessary when the source of truth is the metrics documentation. Also, it's ethically sketchy to directly copy that work.

### DIFF
--- a/pipeline/inputs/nginx.md
+++ b/pipeline/inputs/nginx.md
@@ -1,24 +1,24 @@
-# NGINX Exporter Metrics
+# NGINX Exporter metrics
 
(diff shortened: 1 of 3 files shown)


pipeline/inputs/nginx.md
```diff
@@ -28,7 +28,7 @@ server {
         index  index.html index.htm;
     }
     // configure the stub status handler.
-    location /status {
+    loc.     ation /status {
```
cnorris-cs:
```suggestion
    location /status {
```

PR #1769, opened 2025-06-20
Adding YAML examples to developers wasm filter plugin docs. Fixes #1768.

### LINKED ISSUE #1768: Missing YAML examples for developers wasm filter plugin docs.
The docs for [developers wasm filter plugin](https://docs.fluentbit.io/manual/development/wasm-filter-plugins) are missing YAML examples.

eschabell:
Adding YAML examples to developers wasm filter plugin docs. Fixes #1768.

### DIFF
--- a/development/wasm-filter-plugins.md
+++ b/development/wasm-filter-plugins.md
@@ -98,7 +98,31 @@ pub extern "C" fn rust_filter(tag: *const c_char, tag_len: u32, time_sec: u32, t
 
 The `//export XXX` attribute on TinyGo and `#[no_mangle]` attribute on Rust are required. This is because TinyGo and Rust will mangle their function names if they aren't specified.
 
-Once built, a Wasm program will be available. Then you can execute that built program with the following Fluent Bit configuration:
+Once built, a Wasm program will be available. Then you can execute that built program with one of the following Fluent Bit configurations:
+
+{% tabs %}
+{% tab title="fluent-bit.yaml" %}
+
+```yaml
+pipeline:
+    inputs:
+        - name: dummy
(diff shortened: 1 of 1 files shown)


development/wasm-filter-plugins.md
```diff
@@ -110,13 +134,16 @@ Once built, a Wasm program will be available. Then you can execute that built pr
     Match dummy.*
     WASM_Path /path/to/built_filter.wasm
     Function_Name super_awesome_filter
-    accessible_paths .,/path/to/fluent-bit
+    accessible_paths /path/to/fluent-bit
```
patrick-stephens, line 138: Just checking this change, is the original one a comma separated list with . allowing you to be path agnostic @cosmo0920 or is it just a typo?

cosmo0920: It isn't typo. Just including for the current working directory.
However, this should be removed if we pursue the rigid permission list approach.
Sometimes, I put Wasm programs into /path/to/fluent-bit/build directory and run fluent-bit in /path/to/fluent-bit/build directory. So, if we remove the magic `.` period symbol, we need to write the actual path of the place that contains the actual Wasm programs.

patrick-stephens: Let's add that info to the docs then I think, at least call it out in the example here

eschabell: Ok, so I will add some text to refer to the fact that fluent bit needs to run from the same directory as the '/path/to/built_filter.wasm' file (aka the 'wasm_path').

cosmo0920 (approved):

PR #1841, opened 2025-07-02
Pipeline: input: win event log2-: style

esmerel:


pipeline/inputs/windows-event-log.md
```diff
@@ -33,12 +33,12 @@ Here is a minimum configuration example.
     Match  *
 ```
 
-Note that some Windows Event Log channels (like `Security`) requires an admin privilege for reading. In this case, you need to run fluent-bit as an administrator.
+Some Windows Event Log channels, like `Security` require administrative privileges for reading. In this case, you need to run Fluent Bit as an administrator.
```
cnorris-cs:
```suggestion
Some Windows Event Log channels, like `Security`, require administrative privileges for reading. In this case, you need to run Fluent Bit as an administrator.
```

pipeline/inputs/windows-event-log.md
```diff
@@ -33,12 +33,12 @@ Here is a minimum configuration example.
...
 
-### Command Line
+### Command line
 
-If you want to do a quick test, you can run this plugin from the command line.
+If you want to do a test, you can run this plugin from the command line.
```
cnorris-cs:
```suggestion
If you want to do a test, you can run this plugin from the command line:
```

cnorris-cs (approved):

PR #1878, opened 2025-07-07
pipeline: parsers: decoders: general cleanup

alexakreizinger:
this PR is part of the effort to meet [guidelines](https://github.com/fluent/fluent-bit-docs/blob/master/CONTRIBUTING.md#style-guidelines) for style, clarity, and consistency.


pipeline/parsers/decoders.md
```diff
@@ -16,20 +14,18 @@ The Docker log message encapsulates something like this:
...
-The original message is handled as an escaped string. Fluent Bit wants to use the
-original structured message and not a string.
+The original message is handled as an escaped string. Fluent Bit wants to use the original structured message and not a string.
 
-## Getting Started
+## Get started
 
-Decoders are a built-in feature available through the Parsers file. Each parser
-definition can optionally set one or more decoders. There are two types of decoders:
+Decoders are a built-in feature available through the `parsers.conf` file. Each parser definition can optionally set one or more decoders. There are two types of decoders:
```
alexakreizinger:
```suggestion
Decoders are a built-in feature of parsers in Fluent Bit. Each parser definition can optionally set one or more decoders. There are two types of decoders:
```

pipeline/parsers/decoders.md
```diff
@@ -1,8 +1,6 @@
 # Decoders
 
-There are cases where the log messages being parsed contain encoded data. A typical
-use case can be found in containerized environments with Docker. Docker logs its
-data in JSON format, which uses escaped strings.
+There are cases where the log messages that you want to parse contain encoded data. A typical use case can be found in containerized environments with Docker. Docker logs its data in JSON format, which uses escaped strings.
```
cnorris-cs:
```suggestion
There are cases where the log messages you want to parse contain encoded data. A typical use case can be found in containerized environments with Docker. Docker logs its data in JSON format, which uses escaped strings.
```

pipeline/parsers/decoders.md
```diff
@@ -16,20 +14,18 @@ The Docker log message encapsulates something like this:
 {"log":"{\"status\": \"up and running\"}\r\n","stream":"stdout","time":"2018-03-09T01:01:44.851160855Z"}
 ```
 
-The original message is handled as an escaped string. Fluent Bit wants to use the
-original structured message and not a string.
+The original message is handled as an escaped string. Fluent Bit wants to use the original structured message and not a string.
```
cnorris-cs: Wanting something is what a human would do, and not an application. It can request things. It can do things. But it doesn't want things.

```suggestion
The original message is handled as an escaped string. Fluent Bit will use the original structured message, and not a string.
```

pipeline/parsers/decoders.md
```diff
@@ -16,20 +14,18 @@ The Docker log message encapsulates something like this:
...
 
-## Getting Started
+## Get started
 
-Decoders are a built-in feature available through the Parsers file. Each parser
-definition can optionally set one or more decoders. There are two types of decoders:
+Decoders are a built-in feature of parsers in Fluent Bit. Each parser definition can optionally set one or more decoders. There are two types of decoders:
```
cnorris-cs: You don't have to use this exact suggestion, but I'm trying to get us out of the habit of saying "here are the two types" or "select from these three options", because putting a number in there makes potential technical debt. What if they add another type? Then we need to not only add the info, but to update the intro section for that list. It's a small thing, but it adds up.

```suggestion
Decoders are a built-in feature of parsers in Fluent Bit. Each parser definition can optionally set one or more decoders. Select from one of these decoder types:
```

pipeline/parsers/decoders.md
```diff
@@ -64,34 +60,29 @@ parsers:
...
+| `do_next` | If the decoder succeeded or failed, apply the next decoder in the list for the same field. |
 
 Actions are affected by some restrictions:
 
-- `Decode_Field_As`: If successful, another decoder of the same type and the same
-  field can be applied only if the data continues being an unstructured message (raw text).
-- `Decode_Field`: If successful, can only be applied once for the same field.
-  `Decode`_Field` is intended to decode a structured message.
+- `Decode_Field_As`: If successful, another decoder of the same type and the same field can be applied only if the data continues being an unstructured message (raw text).
+- `Decode_Field`: If successful, can only be applied once for the same field. `Decode_Field` is intended to decode a structured message.
```
cnorris-cs:
```suggestion
- `Decode_Field`: If successful, can be applied only once for the same field. `Decode_Field` is intended to decode a structured message.
```

PR #1892, opened 2025-07-10
Pipeline: output: azure blob: style

esmerel:

### DIFF
--- a/pipeline/outputs/azure_blob.md
+++ b/pipeline/outputs/azure_blob.md
@@ -1,55 +1,53 @@
 ---
-description: Official and Microsoft Certified Azure Storage Blob connector
+description: Microsoft certified Azure Storage Blob connector
 ---
 
 # Azure Blob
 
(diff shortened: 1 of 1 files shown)


pipeline/outputs/azure_blob.md
```diff
@@ -77,33 +75,38 @@ The following configuration example generates a random message with a custom tag
...
 
-![](../../.gitbook/assets/azure_blob.png)
+![Azure Blob](../../.gitbook/assets/azure_blob.png)
 
 ### Configuring and using Azure Emulator: Azurite
 
 #### Install and run Azurite
 
-The quickest way to get started is to install Azurite using npm:
+1. Install Azurite using `npm`:
```
alexakreizinger, line 86:
```suggestion
1. Install Azurite using npm:
```
I don't think we need to use code ticks here since npm is the name of a library and not the name of a variable in this case (my rationale: we wouldn't say "Install using `homebrew`")

esmerel: then I have to update a vale test and bleh
:D

esmerel: oh, nope, my way is right per google style: https://developers.google.com/style/code-in-text#tool-names command line utility names get code ticks

PR #1914, opened 2025-07-14
add vale rule to use gitbook hint syntax

alexakreizinger:
this PR adds a vale suggestion to use GitBook hints rather than ` > Note: lorem ipsum`.


cnorris-cs (approved):
Looks good. Thanks!

PR #1915, opened 2025-07-15
out_bigquery: add documentation for configurable buffer_size option

tsubaron:
Add documentation for the `buffer_size` option in the BigQuery output plugin.

This complements fluent/fluent-bit#10596

### DIFF
--- a/pipeline/outputs/bigquery.md
+++ b/pipeline/outputs/bigquery.md
@@ -45,7 +45,7 @@ Fluent Bit streams data into an existing BigQuery table using a service account 
 | `provider_id` | GCP workload identity provider. Used to construct the full resource name of the identity provider. Currently only AWS accounts are supported. | _none_ |
 | `google_service_account` | The email address of the Google service account to impersonate. The workload identity provider must have permissions to impersonate this service account, and the service account must have permissions to access Google BigQuery resources ( `write` access to tables) | _none_ |
 | `workers` | The number of [workers](../../administration/multithreading.md#outputs) to perform flush operations for this output. | `0` |
-| buffer_size | Sets the upper limit for the HTTP client buffer size (in bytes) when reading API responses from BigQuery. The buffer grows dynamically as needed until it reaches this limit; if a response exceeds this size, any additional data is discarded and a warning is logged. | `4192` |
+| `buffer_size` | Sets the upper limit for the HTTP client buffer size (in bytes) when reading API responses from BigQuery. The buffer grows dynamically as needed until it reaches this limit. If a response exceeds this size, additional data is discarded and a warning is logged. | `4192` |
 
 See Google's [official documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll) for further details.
 


pipeline/outputs/bigquery.md
```diff
@@ -60,6 +60,7 @@ You must configure workload identity federation in GCP before using it with Flue
 | provider\_id | GCP workload identity provider. Used to construct the full resource name of the identity provider. Currently only AWS accounts are supported. |  |
 | google\_service\_account | Email address of the Google service account to impersonate. The workload identity provider must have permissions to impersonate this service account, and the service account must have permissions to access Google BigQuery resources (e.g. `write` access to tables) |  |
 | workers | The number of [workers](../../administration/multithreading.md#outputs) to perform flush operations for this output. | `0` |
+| buffer_size                | Sets the HTTP client buffer size (in bytes) for reading API responses from BigQuery. | `4192` |
```
patrick-stephens: Is this a fixed size, the lower or upper bound?

tsubaron: Thanks for the review!

The `buffer_size` is a fixed-size allocation for the HTTP client buffer. It is not a lower or upper bound. If a response exceeds this size, the plugin logs a warning.

We’ve updated the documentation to clarify this.

pipeline/outputs/bigquery.md
```diff
@@ -60,6 +60,7 @@ You must configure workload identity federation in GCP before using it with Flue
 | provider\_id | GCP workload identity provider. Used to construct the full resource name of the identity provider. Currently only AWS accounts are supported. |  |
 | google\_service\_account | Email address of the Google service account to impersonate. The workload identity provider must have permissions to impersonate this service account, and the service account must have permissions to access Google BigQuery resources (e.g. `write` access to tables) |  |
 | workers | The number of [workers](../../administration/multithreading.md#outputs) to perform flush operations for this output. | `0` |
+| buffer_size | Sets the HTTP client buffer size (in bytes) for reading API responses from BigQuery. This is a fixed-size allocation; responses larger than this size will trigger a warning. | `4192` |
```
patrick-stephens: Well, I presume the warning is that they won't be handled right or is this saying it will auto-grow?

tsubaron: I initially thought the buffer was fixed and allocated based on the provided size, but after checking the internal logic, it turns out the client dynamically increases the buffer as needed.

Here is the part where it grows the buffer, and just below it, you can see that if the allocation fails, a warning is logged and an error is returned:
https://github.com/tsubaron/fluent-bit/blob/out_bigquery-buffer-size-configurable/src/flb_http_client.c#L1470

I will update the documentation to reflect this behavior.

pipeline/outputs/bigquery.md
```diff
@@ -45,6 +45,7 @@ Fluent Bit streams data into an existing BigQuery table using a service account
 | `provider_id` | GCP workload identity provider. Used to construct the full resource name of the identity provider. Currently only AWS accounts are supported. | _none_ |
 | `google_service_account` | The email address of the Google service account to impersonate. The workload identity provider must have permissions to impersonate this service account, and the service account must have permissions to access Google BigQuery resources ( `write` access to tables) | _none_ |
 | `workers` | The number of [workers](../../administration/multithreading.md#outputs) to perform flush operations for this output. | `0` |
+| buffer_size | Sets the upper limit for the HTTP client buffer size (in bytes) when reading API responses from BigQuery. The buffer grows dynamically as needed until it reaches this limit; if a response exceeds this size, any additional data is discarded and a warning is logged. | `4192` |
```
esmerel:
```suggestion
| `buffer_size` | Sets the upper limit for the HTTP client buffer size (in bytes) when reading API responses from BigQuery. The buffer grows dynamically as needed until it reaches this limit. If a response exceeds this size, additional data is discarded and a warning is logged. | `4192` |
```

esmerel (approved):

PR #1919, opened 2025-07-15
Fixes to broken tab constructs on Loki and Kafka output docs. Part of issue 1916.

eschabell:
The construct below had missing first bracket, so fixed for Loki and Kafka docs to get tabs working:

`{% tabs %}`


patrick-stephens (approved):
I'm not a fan of the whack-a-mole so wondering if we can catch this in linting?

PR #1939, opened 2025-07-16
Pipeline: output: flowcounter: style

esmerel:

### DIFF
--- a/pipeline/outputs/flowcounter.md
+++ b/pipeline/outputs/flowcounter.md
@@ -1,6 +1,6 @@
-# FlowCounter
+# Flow counter
 
-The _FlowCounter_ output plugin lets you count up records and their size.
+The _Flow counter_ output plugin lets you count up records and their size.
 
 ## Configuration parameters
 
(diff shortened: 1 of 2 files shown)


pipeline/outputs/flowcounter.md
```diff
@@ -1,29 +1,29 @@
 # FlowCounter
 
-_FlowCounter_ is the protocol to count records. The **flowcounter** output plugin allows to count up records and its size.
+The _FlowCounter_ output plugin lets you count up records and their size.
```
alexakreizinger: do we want to change this to "Flow counter"? I'm not sure why it's pascal case... it doesn't seem like a proper noun :P

esmerel: I was going with it because I couldn't definitively change it to something else but.. I think you're probably right because "Flow counter" is a real thing.

pipeline/outputs/flowcounter.md
```diff
@@ -1,29 +1,29 @@
 # FlowCounter
```
esmerel:
```suggestion
# Flow counter
```

pipeline/outputs/flowcounter.md
esmerel:
```suggestion
The _Flow counter_ output plugin lets you count up records and their size.
```

PR #1953, opened 2025-07-19
pipeline: filters: lookup added new filter page

olegmukhin:
New filter documentation page outlines description, example configuration and CSV handling for the new LookUp filter to be implemented as part #fluent/fluent-bit/pull/10620.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Lookup filter documentation and navigation entry: configuration options with defaults; examples in multiple config formats; sample input/output; metrics (processed/matched/skipped); behavior notes on case sensitivity, header handling, quotes, whitespace trimming, multiline records and data-type handling; clarification that the lookup table is loaded at startup.

### DIFF
--- a/pipeline/filters/lookup.md
+++ b/pipeline/filters/lookup.md
@@ -9,7 +9,7 @@ The plugin supports the following configuration parameters:
 | Key | Description | Default |
 | :-- | :---------- | :------ |
 | `data_source` | Path to the CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. See [Key Considerations](#key-considerations) for details. | _none_ (required) |
-| `lookup_key` | Specifies the record key whose value to search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax for nested fields and array indexing (for example, `$user['profile']['id']`, `$users[0]['id']`). | _none_ (required) |
+| `lookup_key` | Specifies the record key whose value to search for in the CSV file's first column. Supports [record accessor](../../administration/configuring-fluent-bit/classic-mode/record-accessor.md) syntax for nested fields and array indexing (for example, `$user['profile']['id']`, `$users[0]['id']`). | _none_ (required) |
 | `result_key` | If a CSV entry whose value matches the value of `lookup_key` is found, specifies the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found. If this key already exists in the record, it will be overwritten. | _none_ (required) |
 | `ignore_case` | Specifies whether to ignore case when searching for `lookup_key`. If `true`, searches are case-insensitive. If `false`, searches are case-sensitive. Case normalization applies to both the lookup key from the record and the keys in the CSV file. | `false` |
 | `skip_header_row` | If `true`, the filter skips the first row of the CSV file, treating it as a header. If `false`, the first row is processed as data. | `false` |
@@ -142,7 +142,7 @@ Each metric includes a `name` label to identify the filter instance.
 
 ## Key considerations
 
-- The CSV is used to create an in-memory key value lookup table. Column 1 of the CSV is always used as key, while column 2 is assumed to be the value. All other columns in the CSV are ignored.
+- The CSV is used to create an in-memory key-value lookup table. Column 1 of the CSV is always used as key, while column 2 is assumed to be the value. All other columns in the CSV are ignored.
 - CSV fields can be enclosed in double quotes (`"`). Lines with unmatched quotes are logged as warnings and skipped.
 - Multiline values in CSV file aren't currently supported.
 - Duplicate keys (values in first column) in the CSV will use the last occurrence (hash table behavior)


pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
+# Lookup
+
+The Lookup plugin looks up a key value from a record in a specified CSV file and, if a match is found, adds the corresponding value from the CSV as a new key-value pair to the record.
```
alexakreizinger:
```suggestion
The _Lookup_ plugin searches for a specified key in a CSV file and, if it finds a match, adds that key and its corresponding value to the record.
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Configuration parameters
+
+The plugin supports the following configuration parameters
```
alexakreizinger:
```suggestion
The plugin supports the following configuration parameters:
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+
+| Key | Description | Default |
+| :-- | :---------- | :------ |
+| `file` | The CSV file that Fluent Bit will use as a lookup table. The file should contain two columns (key and value), with the first row as an optional header that is skipped. Supports quoted fields and escaped quotes. | _none_ |
```
alexakreizinger:
```suggestion
| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores it. Supported quoted fields and escaped quotes. | _none_ |
```

suggestion for some minor style/clarity stuff, but I just want to double-check that I'm not misunderstanding: does this mean that the filter ALWAYS skips the first row, and therefore users should avoid putting any actual key/values in there? or does fluent bit "know" whether the first row is a header or not?

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+| `lookup_key` | The specific key in the input record to look up in the CSV file's first column. Supports [record accessor](../../administration/configuring-fluent-bit/record-accessor). | _none_ |
```
alexakreizinger:
```suggestion
| `lookup_key` | Specifies the key to search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax. | _none_ |
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+| `result_key` | The name of the key to add to the output record with the matched value from the CSV file's second column if a match is found. | _none_ |
```
alexakreizinger:
```suggestion
| `result_key` | If `lookup_key` is found, sets the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
```

(hopefully this still makes sense... if not, let me know! 😅)

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+| `ignore_case` | Ignore case when matching the lookup key against the CSV keys. | `false` |
```
alexakreizinger:
```suggestion
| `ignore_case` | Specifies whether to ignore case when searching for `lookup_key`. If `true`, searches are case-insensitive. If `false`, searches are case-sensitive. Possible values: `true`, `false`. | `false` |
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+
+[OUTPUT]
+    Name   stdout
+    Match  test
+```
+
+{% endtab %}
+{% endtabs %}
+
+The following configuration reads log records from `devices.log` that includes the following values for device hostnames:
```
alexakreizinger:
```suggestion
The previous configuration reads log records from `devices.log` that includes the following values for device hostnames:
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+{"hostname": " host with space "}
+{"hostname": "quoted \"host\""}
+{"hostname": "unknown-host"}
+{}
+{"hostname": [1,2,3]}
+{"hostname": {"sub": "val"}}
+{"hostname": " "}
+```
+
+It uses the value of the `hostname` field (which has been set as the `lookup_key`) to find matching values in column 1 of the  (`device-bu.csv`) CSV file.
```
alexakreizinger:
```suggestion
It uses the value of the `hostname` field (which was set as the `lookup_key`) to find matching values in the first column of `device-bu.csv`.
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+db-test-abc,Marketing
+web-frontend-xyz,Marketing
+app-backend-123,Operations
+"legacy-system true","Legacy IT"
+" host with space ","Infrastructure"
+"quoted ""host""", "R&D"
+no-match-host,Should Not Appear
+```
+
+Where a match is found the filter adds new key (name of which is set by the `result_key` input) with the value from the second column of the CSV file of the matched row.
```
alexakreizinger:
```suggestion
When the filter finds a match, it adds a new key with the name specified by `result_key` and a value from the second column of the CSV file of the row where `lookup_key` was found.
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+
+For above configuration the following output can be expected (when matching case is ignored as `ignore_case` is set to true):
```
alexakreizinger:
```suggestion
This results in output that resembles the following:
```

pipeline/filters/lookup.md
olegmukhin: The filter always skips the first row. The file can have more than 2 columns, but all other columns will be ignored and not loaded (i.e. it will not cause an error).

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,134 @@
...
+{"hostname"=>{"sub"=>"val"}}
+```
+
+## CSV import
+
+The CSV is used to create an in-memory key value lookup table. Column 1 of the CSV is always used as key, while column 2 is assumed to be the value. All other columns in the CSV are ignored.
+
+This filter is intended for static datasets. CSV is loaded once when Fluent Bit starts and is not reloaded.
+
+Multiline values in CSV file are not currently supported.
```
alexakreizinger:
```suggestion

Fluent Bit creates an in-memory key/value lookup table from the CSV file that you provide. The first column of this CSV is always treated as a key, and its second column as a value. Any other columns are ignored.

This filter is intended for static datasets. After Fluent Bit loads the CSV file, it won't reload that file, which means the filter's lookup table won't update to reflect any changes.

This filter doesn't support multiline values in CSV files.
```

pipeline/filters/lookup.md
olegmukhin: This might be a little confusing - it does not use the key from CSV file - it uses the result_key input as the key and the CSV value (i.e. second column) as the value and adds that combination to the record.

Thank you for all the clarifications.

alexakreizinger: ah, that makes sense! I'll clarify some of those suggestions accordingly...

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
+# Lookup
+
+The Lookup plugin looks up a key value from a record in a specified CSV file and, if a match is found, adds the corresponding value from the CSV as a new key-value pair to the record.
```
alexakreizinger:
```suggestion
The _Lookup_ plugin searches for a specified key in a CSV file and, if it finds a match, adds a new key to the output record whose value matches the corresponding CSV value.
```
hopefully this is better... it doesn't get into the specifics of `result_key` quite yet, but I think it's okay to save that for the following section with more details

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Configuration parameters
+
+The plugin supports the following configuration parameters:
+
+| Key | Description | Default |
+| :-- | :---------- | :------ |
+| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores it. Supported quoted fields and escaped quotes. | _none_ |
+| `lookup_key` | Specifies the key to search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax. | _none_ |
+| `result_key` | If `lookup_key` is found, sets the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
```
alexakreizinger:
```suggestion
| `result_key` | If `lookup_key` is found, specifies the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
```
(slightly tweaking my own suggestion, lol)

pipeline/filters/lookup.md
olegmukhin: The ignore_case is only applicable when matching the lookup_key value to the first column of the CSV (which is the key for the CSV). I think the change here can be interpreted that ignore_case is for how the lookup_key is matched against record rather than CSV. It's quite a tricky thing to describe actually, but flow is lookup_key defines a record assessor that will be used to find a value in the record; that value is then matched against the first column of the CSV; if a match is found the corresponding value for the matched row is taken from the second column; this extracted value from the second column is then added to the record under a key defined under result_key. Hope this helps :)

alexakreizinger: good callout... let me chew on that for a minute and see if I can suggest something :P

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+## Configuration parameters
+
+The plugin supports the following configuration parameters:
+
+| Key | Description | Default |
+| :-- | :---------- | :------ |
+| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores it. Supported quoted fields and escaped quotes. | _none_ |
+| `lookup_key` | Specifies the key to search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax. | _none_ |
+| `result_key` | If `lookup_key` is found, specifies the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
+| `ignore_case` | Ignore case when matching the lookup key against the CSV keys. | `false` |
```
alexakreizinger:
```suggestion
| `ignore_case` | Specifies whether to ignore case when searching for the value of `lookup_key` in the specified CSV file. If `true`, searches are case-insensitive. If `false`, searches are case-sensitive. Possible values: `true`, `false`. | `false` |
```
apologies for the delay... so I think I may have fundamentally misunderstood how this filter works, which means I should update my other suggestions as well, lol. it this specific suggestion more accurate?

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+
+[OUTPUT]
+    Name   stdout
+    Match  test
+```
+
+{% endtab %}
+{% endtabs %}
+
+The previous configuration reads log records from `devices.log` that includes the following values for device hostnames:
```
alexakreizinger:
```suggestion
The previous configuration reads log records from `devices.log` that includes the following values in the `hostname` field:
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+{"hostname": " host with space "}
+{"hostname": "quoted \"host\""}
+{"hostname": "unknown-host"}
+{}
+{"hostname": [1,2,3]}
+{"hostname": {"sub": "val"}}
+{"hostname": " "}
+```
+
+It uses the value of the `hostname` field (which was set as the `lookup_key`) to find matching values in the first column of `device-bu.csv`.
```
alexakreizinger:
```suggestion
Because `hostname` was set as the `lookup_key`, the Lookup filter uses the value of each `hostname` key within the record to search for matching values in the first column of the CSV file.
```

pipeline/filters/lookup.md
alexakreizinger:
```suggestion
The _Lookup_ plugin searches a CSV file for values that match the value of a specific key in a record and then, if it finds a match, adds a new key to the record whose value matches the corresponding CSV value.
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+The Lookup plugin looks up a key value from a record in a specified CSV file and, if a match is found, adds the corresponding value from the CSV as a new key-value pair to the record.
+
+## Configuration parameters
+
+The plugin supports the following configuration parameters:
+
+| Key | Description | Default |
+| :-- | :---------- | :------ |
+| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores it. Supported quoted fields and escaped quotes. | _none_ |
+| `lookup_key` | Specifies the key to search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax. | _none_ |
```
alexakreizinger:
```suggestion
| `lookup_key` | Specifies the record key whose value search for in the CSV file's first column. Supports [record accessor](../administration/configuring-fluent-bit/classic-mode/record-accessor) syntax. | _none_ |
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+| `result_key` | If `lookup_key` is found, specifies the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
```
alexakreizinger:
```suggestion
| `result_key` | If a CSV entry whose value matches the value of `lookup_key` is found, specifies the name of the new key to add to the output record. This new key uses the corresponding value from the second column of the CSV file in the same row where `lookup_key` was found.  | _none_ |
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+
+The Lookup plugin looks up a key value from a record in a specified CSV file and, if a match is found, adds the corresponding value from the CSV as a new key-value pair to the record.
+
+## Configuration parameters
+
+The plugin supports the following configuration parameters:
+
+| Key | Description | Default |
+| :-- | :---------- | :------ |
+| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores it. Supported quoted fields and escaped quotes. | _none_ |
```
alexakreizinger:
```suggestion
| `file` | The CSV file that the Lookup filter will use as a lookup table. This file must contain one column of keys and one column of values. The Lookup filter treats the first row as a header and ignores these values. Supports quoted fields and escaped quotes. | _none_ |
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+{}
+{"hostname": [1,2,3]}
+{"hostname": {"sub": "val"}}
+{"hostname": " "}
+```
+
+It uses the value of the `hostname` field (which was set as the `lookup_key`) to find matching values in the first column of `device-bu.csv`.
+
+```text
+hostname,business_line
```
alexakreizinger, line 101:
```suggestion
host,department
```
minor suggestion, but I wonder if it would make the examples clearer to use values for the CSV header row that are NOT found anywhere else. that way it doesn't seem like `hostname` or `business_line` are referring to the CSV header row items

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+    Name   stdout
+    Match  test
+```
+
+{% endtab %}
+{% endtabs %}
+
+The previous configuration reads log records from `devices.log` that includes the following values for device hostnames:
+
+```text
```
alexakreizinger, line 83:
```suggestion
```shell
```

pipeline/filters/lookup.md
```diff
@@ -0,0 +1,135 @@
...
+" host with space ","Infrastructure"
+"quoted ""host""", "R&D"
+no-match-host,Should Not Appear
+```
+
+When the filter finds a match, it adds a new key with the name specified by `result_key` and a value from the second column of the CSV file of the row where `lookup_key` was found.
+
+This results in output that resembles the following:
+
+```text
```
alexakreizinger, line 117:
````suggestion
```shell
````

alexakreizinger (approved):
thanks for incorporating those changes @olegmukhin (and thank you for opening a PR in the first place—I meant to say that earlier but forgot 😅)

I'm approving this but I'll let you merge it... not sure if we're waiting on corresponding code changes before the docs are ready to go live :)

PR #1955, opened 2025-07-21
fix(install): update Ubuntu installation commands for non-root users

mohamedabdallah20:
- Modified GPG key import command to use `sudo sh -c` for proper permission handling
- Simplified repository setup into a single pipe-to-tee command to avoid manual file editing
- Ensures commands work safely without requiring direct root shell access


patrick-stephens (approved):
Thanks, any chance we can update the other linux targets as well?

mohamedabdallah20:
@patrick-stephens Sure

patrick-stephens (approved):

Issue #1957, opened 2025-07-21
Administration YAML service section docs have broken links for hot reload docs.


eschabell:
Link to detailed hot reload documentation found in the [service section configuration doc](https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/yaml/service-section) is broken.

PR #1959, opened 2025-07-21
Fixing links from hot reload, broken due to underscore in file name changed to dash in link. Part of issue #1957.

eschabell:
Fixing links from hot reload, broken due to underscore in file name changed to dash in link. Part of issue #1957.


esmerel (approved):
FYI, the markdown tables should work fine without all the extra spacing.

update looks good

PR #1966, opened 2025-07-22
Pipeline: outputs: cloudwatch: fix `auto_create_group` value

andylim0221:
Fixed the `auto_create_group` value in CloudWatch output config


cosmo0920 (changes requested):
Fluent Bit also accepts On as true. So, it's not changed the meanings.
If we used On as meant for true, we shouldn't change from On.

eschabell:
@alexakreizinger can you review before I look to merge this please?

PR #2003, opened 2025-07-27
Update installation on ubuntu commands

mohamedabdallah20:
remove $CODENAME that doesn't work properly

### DIFF
--- a/installation/linux/ubuntu.md
+++ b/installation/linux/ubuntu.md
@@ -45,10 +45,16 @@ Refer to the [supported platform documentation](../supported-platforms.md) to se
 
 ## Update your sources lists
 
-On Ubuntu, you need to add the Fluent Bit APT server entry to your sources lists. Ensure `CODENAME` is set to your specific [Ubuntu release name](https://wiki.ubuntu.com/Releases). For example, `focal` for Ubuntu 20.04.
+On Ubuntu, you need to add the Fluent Bit APT server entry to your sources lists.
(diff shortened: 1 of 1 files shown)


installation/linux/ubuntu.md
```diff
@@ -48,7 +48,7 @@ Refer to the [supported platform documentation](../supported-platforms.md) to se
 On Ubuntu, you need to add the Fluent Bit APT server entry to your sources lists. Ensure `CODENAME` is set to your specific [Ubuntu release name](https://wiki.ubuntu.com/Releases). For example, `focal` for Ubuntu 20.04.
 
 ```bash
-echo "deb [signed-by=/usr/share/keyrings/fluentbit-keyring.gpg] https://packages.fluentbit.io/ubuntu/${CODENAME} ${CODENAME} main" | sudo tee /etc/apt/sources.list.d/fluent-bit.list
+echo "deb [signed-by=/usr/share/keyrings/fluentbit-keyring.gpg] https://packages.fluentbit.io/ubuntu/$(grep -oP '(?<=VERSION_CODENAME=).*' /etc/os-release || lsb_release -cs) $(grep -oP '(?<=VERSION_CODENAME=).*' /etc/os-release || lsb_release -cs) main" | sudo tee /etc/apt/sources.list.d/fluent-bit.list
```
patrick-stephens: I think we need to redirect `stderr` for `lsb_release` messages, otherwise you can get extra lines of output like `No LSB modules are available.`

PR #2011, opened 2025-07-28
Admin: config: classic: commands: final style

esmerel:
Last round of cleanup for vale and markdown stuff


cnorris-cs (approved):
Looks good. Approving to move this forward.

PR #2025, opened 2025-07-30
Vale: raising alert level of some tests

esmerel:

This PR updates several vale tests from suggestion to warning or error. It also corrects a few files where errors would be flagged with the rule level changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,13 +140,17 @@ The following tests are at a `error` level and will prevent merging:
 
 - [AmSpelling](https://developers.google.com/style/word-list)
 - [Ampersand](https://developers.google.com/style/word-list#ampersand)
+- Don't use
 - [Emdash](https://github.com/errata-ai/Google/blob/master/Google/EmDash.yml)
 - [Endash](https://github.com/errata-ai/Google/blob/master/Google/EmDash.yml)
 - [Exclamation](https://github.com/errata-ai/Google/blob/master/Google/Exclamation.yml)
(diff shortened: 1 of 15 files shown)


CONTRIBUTING.md
```diff
@@ -42,7 +42,7 @@ within a few minutes.
 If you open a pull request that requires ongoing discussion or review, the Fluent Bit
 maintainers will add a `waiting-for-user` to your pull request. This tag means that
 we're blocked from moving forward until you reply. To keep contributions from going
-stale, we'll wait 45 days for your response, but we may close the pull request if we
+stale, we'll wait 45 days for your response, but we might close the pull request if we
```
alexakreizinger, line 45: would it make sense to say "we reserve the right to close your pull request if we don't hear back from you by then"?

CONTRIBUTING.md
```diff
@@ -140,13 +140,17 @@ The following tests are at a `error` level and will prevent merging:
 
 - [AmSpelling](https://developers.google.com/style/word-list)
 - [Ampersand](https://developers.google.com/style/word-list#ampersand)
+- Don't use
```
alexakreizinger:
```suggestion
- [Don'tUse](https://github.com/fluent/fluent-bit-docs/blob/master/vale-styles/FluentBit/DontUse.yml)
```

alexakreizinger (approved):

CONTRIBUTING.md
esmerel: I have a ticket to rework this page so we can evaluate this thought when we do that. It should be more direct for sure.

PR #2035, opened 2025-08-05
Update Log Analytics plugin docs for OCI service rename

rghouzra:
This pull request updates the documentation for the Oracle Cloud Infrastructure Logging Analytics output plugin to reflect recent service name change at OCI from Logging Analytics to Log Analytics. The plugin name itself is not being changed to not introduce any breaking changes.


esmerel (approved):
Awesome, thanks for the update!

PR #2048, opened 2025-08-22
Add fluentbit_input_ingestion_paused metric to the documentation

antonio:
:wave:

While working on reducing dropped logs in a high-volume Kubernetes setup, we found the ingestion paused input metric (added in https://github.com/fluent/fluent-bit/pull/8044) to be super helpful for validating Fluent Bit health under pressure and for tuning settings like `mem_buf_limit`. We realized it was not called out in the docs, so this PR adds it to the metrics reference for future operators.


cosmo0920 (approved):
Yeah, it's a missing piece of our docs to be one of the rigorous indicators to indicate saturating for pipeline operations. Good catch!

esmerel:
If @cosmo0920 is ok with it then I'll go ahead and merge it. Thanks @antonio!

PR #2072, opened 2025-09-25
intallation: Abstracting duplicated info

esmerel:

This abstracts some duplicated info out of some of the installation files and crosslinks rather than duplicates.


alexakreizinger (approved):
LGTM - only question/suggestion is whether (in the areas where you removed all of those duplicated single-line install instructions) it would be helpful to add a link back to the `downloads/linux##single-line-install` header so people can quickly jump to that info

cosmo0920 (approved):

PR #2095, opened 2025-10-07
multiline: Add a description for the buffer limit of multiline

cosmo0920:
This is corresponding to https://github.com/fluent/fluent-bit/pull/10653.


esmerel (approved):
Looks good to me

administration/configuring-fluent-bit/multiline-parsing.md
```diff
@@ -255,6 +255,23 @@ The multiline parser is a very powerful feature, but it has some limitations tha
...
+| ------------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `multiline_buffer_limit` | Sets the maximum size of the in-memory buffer used while assembling a multiline message. When the accumulated size exceeds this limit, the message is truncated and a `multiline_truncated: true` metadata field is attached to the emitted record. A value of `0` disables the limit. Accepts unit suffixes like `KiB`, `MiB`, or `GiB`. | `2MiB`  |
+
+```text
+service:
+  multiline_buffer_limit  2MiB   # default; limit concatenated multiline message size
+```
+
+If the limit is reached, Fluent Bit flushes the partial record with `multiline_truncated: true` metadata immediately to prevent Out-Of-Memory (OOM) conditions.
```
lecaros, line 271: does this limit work as a soft-limit for the current record being appended? or is the ml record flushed without appending the current record?
what happens to the remaining records intended to be appended to the ml record? are they discarded or flushed as individual records?

cosmo0920: > does this limit work as a soft-limit for the current record being appended? or is the ml record flushed without appending the current record?
> what happens to the remaining records intended to be appended to the ml record? are they discarded or flushed as individual records?

This works as a hard limit and the remaining records will not be concatenated.
They will be flush as individual record(s).

Issue #2096, opened 2025-10-07
YAML  introduction docs update (user suggestion)


eschabell:
Raising issue here to capture user suggestion from online forum: https://fluent-all.slack.com/archives/C0CTQGHKJ/p1759792361290489?thread_ts=1759750481.930109&cid=C0CTQGHKJ

_**Suggested change:**

ref: https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/yaml

**Before you get started**
YAML has become essential in a cloud ecosystem. To minimize friction and provide a more intuitive experience for creating data pipelines, users are encouraged to transition to YAML. The YAML format enables features, such as processors, that aren't possible to configure in classic mode.
NOTE:  Fluent Bit traditionally offered a classic configuration mode, a custom configuration format that's phasing out. While classic mode has served well for many years, it has several limitations. Its basic design only supports grouping sections with key-value pairs and lacks the ability to handle sub-sections or complex data structures like lists._

PR #2097, opened 2025-10-07
Config: classic mode: adding index page

esmerel:

Turns the "readme" file into a real page for classic config mode and rewords the yaml config page a little based on a suggestion from #2096

### DIFF
--- a/administration/configuring-fluent-bit/classic-mode.md
+++ b/administration/configuring-fluent-bit/classic-mode.md
@@ -2,9 +2,7 @@
 
 Fluent Bit classic mode configuration will be deprecated at the end of 2026.
 
-Classic mode is a custom configuration model for Fluent Bit. It's more limited than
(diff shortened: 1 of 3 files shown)


administration/configuring-fluent-bit/classic-mode.md
```diff
@@ -0,0 +1,14 @@
+# Configure Fluent Bit using classic mode
+
+Fluent Bit classic mode configuration will be deprecated at the end of 2026.
```
alexakreizinger, line 4: would it be overkill to put this in an alert box?

```suggestion
{% hint style="info" %}
Fluent Bit classic mode configuration will be deprecated at the end of 2026.
{% endhint %}
```

esmerel: No it would not. It wasn't on the other page and I just copied it and didn't think about offsetting it :D

PR #2133, opened 2025-11-04
fix confusing/misleading info about parser support in YAML config

alexakreizinger:
bear with me... this is going to be the first PR in a series of PRs trying to clarify some confusing stuff related to parsers, processors, and other YAML configuration-related topics. but I'm starting with this because it was the tiny thread that kept unraveling the more I pulled at it.

long story short, I thought it was odd that one of the configuration file pages stated that YAML configuration files don't support parsers. then I realized that it's because YAML configuration files don't have a _dedicated parsers section_—which is to say, that parsers are defined on a per-plugin basis.

so this PR attempts to perform the minimum viable amount of cleanup to clear up that confusion. that said, I'm going to need to revisit some of the docs in this section to cut through other things that are potentially misleading and to consolidate duplicated/scattered info.

### DIFF
--- a/administration/configuring-fluent-bit/yaml/configuration-file.md
+++ b/administration/configuring-fluent-bit/yaml/configuration-file.md
@@ -10,12 +10,13 @@ src="https://static.scarf.sh/a.png?x-pxid=864c6f0e-8977-4838-8772-84416943548e" 
 One of the ways to configure Fluent Bit is using a YAML configuration file that works at a global scope. These YAML configuration files support the following top-level sections:
 
 - `env`: Configures [environment variables](../administration/configuring-fluent-bit/yaml/environment-variables-section).
-- `includes`: Specifies additional YAML configuration files to [include as part of a parent file](../manual/administration/configuring-fluent-bit/yaml/includes-section).
-- `service`: Configures global properties of the Fluent Bit [service](../manual/administration/configuring-fluent-bit/yaml/service-section).
-- `pipeline`: Configures active [`inputs`, `filters`, and `outputs`](/manual/administration/configuring-fluent-bit/yaml/pipeline-section).
+- `includes`: Specifies additional YAML configuration files to [include as part of a parent file](../administration/configuring-fluent-bit/yaml/includes-section).
+- `service`: Configures global properties of the Fluent Bit [service](../administration/configuring-fluent-bit/yaml/service-section).
+- `pipeline`: Configures active [`inputs`, `filters`, and `outputs`](../administration/configuring-fluent-bit/yaml/pipeline-section).
 - `parsers`: Defines [custom parsers](../administration/configuring-fluent-bit/yaml/parsers-section).
-- `multiline_parsers`: Defines [custom multiline parsers](../manual/administration/configuring-fluent-bit/yaml/multiline-parsers-section).
-- `plugins`: Defines paths for [custom plugins](/manual/administration/configuring-fluent-bit/yaml/plugins-section).
+- `multiline_parsers`: Defines [custom multiline parsers](../administration/configuring-fluent-bit/yaml/multiline-parsers-section).
+- `plugins`: Defines paths for [custom plugins](../administration/configuring-fluent-bit/yaml/plugins-section).
+- `upstream_servers`: Defines [nodes](../administration/configuring-fluent-bit/yaml/upstream-servers-section) for output plugins.
 
 {% hint style="info" %}
 YAML configuration is used in the smoke tests for containers. An always-correct up-to-date example is here: <https://github.com/fluent/fluent-bit/blob/master/packaging/testing/smoke/container/fluent-bit.yaml>.


administration/configuring-fluent-bit/yaml/configuration-file.md
```diff
@@ -7,21 +7,15 @@ description: Learn about the YAML configuration file used by Fluent Bit
...
-The YAML configuration file doesn't support the following sections:
-
-- `Parsers`
+- `env`
+- `includes`
+- `service`
+- `pipeline`
+  - `inputs`
+  - `outputs`
+  - `filters`
```
cosmo0920: parsers can handle in YAML type of configurations:

This is an example configuration of YAML but it's still valid for the format:

```yaml
env:
  FLB_ENV: "1"

parsers:
  - name: json
    format: json

plugins:
  - /opt/fluent-bit/plugins/custom.so

service:
  flush: 1
  log_level: info

pipeline:
  inputs:
    - name: tail
      path: /var/log/syslog
  outputs:
    - name: stdout

```

cosmo0920 (commented):
I added a comment for the top level elements of YAML configuration.

administration/configuring-fluent-bit/yaml/configuration-file.md
alexakreizinger: 🤔 what does an example configuration like this do? does it apply a JSON parser to all inputs?

administration/configuring-fluent-bit/yaml/pipeline-section.md
```diff
@@ -1,22 +1,21 @@
 # Pipeline
 
-The `pipeline` section defines the flow of how data is collected, processed, and sent to its final destination. It encompasses the following core concepts:
+The `pipeline` section defines the flow of how data is collected, processed, and sent to its final destination. This section contains three subsections:
```
esmerel:
```suggestion
The `pipeline` section defines the flow of how data is collected, processed, and sent to its final destination. This section contains the following subsections:
```

administration/configuring-fluent-bit/yaml/configuration-file.md
cosmo0920: I mean, `parsers`, `multiline_parsers`, and `plugins` are able to be handled in the same YAML configurations.
But in classix format, `parsers`, `multiline_parsers`, and `plugins` should be handled with `parsers_file` and `plugins_file` and loaded as separated other files.
This could be an advantage of YAML format of Fluent BIt configurations.

alexakreizinger: ah, that makes sense :) so those are optional sections you can use to define custom parsers/plugins/etc

of the other sections that you can include in a YAML config, are any of them required? I'm wondering if it would be helpful to specify which ones you always need to include versus which ones are optional.

cosmo0920: The suggested sections are all of optional. So, it's not needed to put in every configuration.

administration/configuring-fluent-bit/yaml/pipeline-section.md
```diff
@@ -1,22 +1,21 @@
 # Pipeline
 
-The `pipeline` section defines the flow of how data is collected, processed, and sent to its final destination. It encompasses the following core concepts:
+The `pipeline` section defines the flow of how data is collected, processed, and sent to its final destination. This section contains the following subsections:
 
 | Name | Description |
 | ---- | ----------- |
 | `inputs` | Specifies the name of the plugin responsible for collecting or receiving data. This component serves as the data source in the pipeline. Examples of input plugins include `tail`, `http`, and `random`. |
-| `processors` | **Unique to YAML configuration**, processors are specialized plugins that handle data processing directly attached to input plugins. Unlike filters, processors aren't dependent on tag or matching rules. Instead, they work closely with the input to modify or enrich the data before it reaches the filtering or output stages. Processors are defined within an input plugin section. |
```
patrick-stephens, line 8: I'm not sure I follow the reasoning around removing this.

alexakreizinger: it's part of a larger change that I'm planning to make in the near future, but TLDR is that I think it's misleading to include processors on the same "level" as inputs/filters/outputs since processors are nested one level below them; this table layout makes them all appear to be on the same hierarchal level. when I revisit this page (and others) I'm hoping to make it clearer exactly how processors get attached to inputs and outputs, and that that they're not a direct parallel to filters.

administration/configuring-fluent-bit/yaml/configuration-file.md
```diff
@@ -7,21 +7,15 @@ description: Learn about the YAML configuration file used by Fluent Bit
...
-The YAML configuration file doesn't support the following sections:
-
-- `Parsers`
+- `env`: Configures [environment variables](../administration/configuring-fluent-bit/yaml/environment-variables-section).
+- `includes`: Specifies additional YAML configuration files to [include as part of a parent file](../manual/administration/configuring-fluent-bit/yaml/includes-section).
+- `service`: Configures global properties of the Fluent Bit [service](../manual/administration/configuring-fluent-bit/yaml/service-section).
+- `pipeline`: Configures active [`inputs`, `filters`, and `outputs`.](/manual/administration/configuring-fluent-bit/yaml/pipeline-section).
+- `parsers`: Defines [custom parsers](../administration/configuring-fluent-bit/yaml/parsers-section).
+- `multiline_parsers`: Defines [custom multiline parsers](../manual/administration/configuring-fluent-bit/yaml/multiline-parsers-section).
+- `plugins`: Defines paths for [custom plugins](/manual/administration/configuring-fluent-bit/yaml/plugins-section).
```
alexakreizinger: @esmerel @cosmo0920 what do you think of this change? (this is also a semi-temporary change, because I want to expand on this information when I revisit this page and section, but hopefully it's okay for now.)

cosmo0920: Got it. It's OK for now.

PR #2145, opened 2025-11-05
Completed configuration options for source build doc. Fixes #2134.

### LINKED ISSUE #2134: Extend list of compilation options for source builds
The list of compilation options is not up to date in the DOC. I'm in the process of making a PR for an update, but I have the impression that certain variables are no longer useful. Here is a list:
FLB_FILTER_ALTER_SIZE
FLB_FILTER_THROTTLE_SIZE
FLB_FILTER_LUA_USE_MPACK

It would be good to add in the pipelines a consistency check between the variables here:

- https://github.com/fluent/fluent-bit/blob/master/CMakeLists.txt
- https://github.com/fluent/fluent-bit/blob/master/cmake/plugins_options.cmake
- https://github.com/fluent/fluent-bit-docs/blob/master/installation/sources/build-and-install.md

_Note that this was originally raised by @tguenneguez and addressed with doc PR https://github.com/fluent/fluent-bit-docs/pull/1386, but aged out of making sense to fix all the conflicts, so starting fresh here._

eschabell:
Completed configuration options for source build doc. Fixes #2134.

**Note to reviewers:** 
The file with changes contains links to docs for the plugins listed, with a few using yet to be made docs. Each of these docs has an open issue and i plan to work through them next. I would like these incorrect (not available) links to be in place now for the fixes I'm rolling out next to save time revisiting them.

### DIFF
--- a/installation/downloads/source/build-and-install.md
+++ b/installation/downloads/source/build-and-install.md
@@ -293,7 +293,7 @@ The following table describes the output plugins available:
 | [`FLB_OUT_LOKI`](../../../pipeline/outputs/loki.md)                                      | Enable Loki output plugin                                | `On`    |
 | [`FLB_OUT_NATS`](../../../pipeline/outputs/nats.md)                                      | Enable NATS output plugin                                | `On`    |
 | [`FLB_OUT_NRLOGS`](../../../pipeline/outputs/new-relic.md)                               | Enable New Relic output plugin                           | `On`    |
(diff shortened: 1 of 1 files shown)


patrick-stephens (approved):
I am of the opinion that this looks good and we should merge now then just resolve any issues raised directly.

One thing we should indicate is the implicit disabling of requested options when dependencies are missing - this is not consistent and can break things without you realising.

PR #2178, opened 2025-11-17
move/consolidate installation requirements

alexakreizinger:
this PR moves the general installation requirements into the doc for building from source, since it turns out that info is _only_ applicable to building from source.

...however, there was already a requirements section in that doc, with slightly different info. so I combined those + added some info per one of our Chronosphere developers who shared what packages he needed to install on a fresh Ubuntu installation.

## Summary by CodeRabbit

* **Documentation**
  * Removed the top-level "Requirements" subsection to reduce duplication.
  * Moved prerequisite details into the dedicated build-and-install section for clarity.
  * Replaced vague prerequisites with an explicit, actionable list of packages and minimum version guidance for building from source.
  * Clarified that some optional input/output plugins may require additional third-party components (e.g., Kafka).

### DIFF
--- a/installation/downloads/source/build-and-install.md
+++ b/installation/downloads/source/build-and-install.md
@@ -8,7 +8,7 @@ To build and install Fluent Bit from source, you must also install the following
 
 - `bison`
 - `build-essentials`
-- `cmake`
+- `cmake` (version 3.31.6 or later)
 - `flex`
 - `libssl-dev`
 - `libyaml-dev`


installation/downloads/source/build-and-install.md
```diff
@@ -4,11 +4,18 @@
 
 ## Requirements
 
-- CMake 3.12 or greater. You might need to use `cmake3` instead of `cmake`.
-- Flex
-- Bison 3 or greater
-- YAML headers
-- OpenSSL headers
+To build and install Fluent Bit from source, you must also install the following packages:
```
cosmo0920, line 7: We need to clarify the minimum requirement of CMake version.
Currently, we use CMake 3.31.6 in our CI.
Plus, we're still supporting to compile on CentOS7.
We also use CMake 3.31.6 for that platform in CI.
So, the minimum requirements of CMake version could be become CMake 3.31.6 now.

installation/downloads/source/build-and-install.md
```diff
@@ -4,11 +4,18 @@
...
+
+- `bison`
+- `build-essentials`
+- `cmake`
+- `flex`
+- `libfl-dev`
```
cosmo0920: There is no libfl library. We use cfl for one of the core Fluent Bit libraries but this is not a part of the distribution repo. So, we don't need to list it here.

cosmo0920 (changes requested):
I found some of the glitches in the updated contents. Could you take a look?

installation/downloads/source/build-and-install.md
alexakreizinger: I can add that :) similarly, do you think we need to add minimum versions for any of these other libraries?

installation/downloads/source/build-and-install.md
alexakreizinger:
```suggestion
```

installation/downloads/source/build-and-install.md
```diff
@@ -4,11 +4,18 @@
...
-- CMake 3.12 or greater. You might need to use `cmake3` instead of `cmake`.
-- Flex
-- Bison 3 or greater
-- YAML headers
-- OpenSSL headers
+To build and install Fluent Bit from source, you must also install the following packages:
+
+- `bison`
+- `build-essentials`
+- `cmake`
```
alexakreizinger:
```suggestion
- `cmake` (version 3.31.6 or greater)
```

cnorris-cs: The Google developer style guide would have us use "later" for version numbers.

```suggestion
- `cmake` (version 3.31.6 or later)
```

installation/downloads/source/build-and-install.md
cosmo0920: I don't think so. We only need to add the minimum required version for CMake for now.

PR #2263, opened 2025-12-02
pipeline: router: Only logs are currently supported

lecaros:
## Summary by CodeRabbit

* **Documentation**
  * Conditional routing docs now show a "Supported" column indicating which signal types (logs, metrics, traces, any) are supported.
  * Removed several example routing sections and example configurations that demonstrated per-signal and multi-signal routing, simplifying the guidance and examples shown.


pipeline/router.md
```diff
@@ -284,12 +284,12 @@ The following comparison operators are available for condition rules:
...
-| `logs` | Routes log records |
-| `metrics` | Routes metric records |
-| `traces` | Routes trace records |
-| `any` | Routes all signal types |
+| Signal | Description | Supported |
+| --- | --- |---|
+| `logs` | Routes log records | Yes |
+| `metrics` | Routes metric records | No | 
+| `traces` | Routes trace records | No |
+| `any` | Routes all signal types | No |
```
alexakreizinger, line 292: I'm still a little bit confused by this--would it help to change the sentence before the table to say something like "Not all signal types support conditional routing"? (right now the way it's phrased makes it sound like something else entirely)

alexakreizinger, line 292: should `any` be part of this table? (does `any` refer to a specific tag/label/name of something? if not, I think it makes more sense to remove it)

PR #2289, opened 2025-12-10
Note that envvar cannot be used in include section

jimin9038:
This PR is a short addition noting that **environment variables cannot be used in the include section**.
Some users may attempt to use environment variables in the include section and eventually run into issues - often ending up at workaround discussions such as this one: https://github.com/fluent/fluent-bit/issues/2020

It would be great to add a note clarifying that the include section does not parse environment variables, and, if possible, include a link to the above issue as a reference for the workaround.

Thank you! 

## Summary by CodeRabbit

* **Documentation**
  * Clarified that environment variables are not supported in include directives.
  * Noted that included file paths must be specified as literal strings.
  * Updated guidance to prevent confusion when referencing external configuration files, ensuring users supply explicit file paths rather than relying on variable expansion.

### DIFF
--- a/administration/configuring-fluent-bit/classic-mode/configuration-file.md
+++ b/administration/configuring-fluent-bit/classic-mode/configuration-file.md
@@ -159,3 +159,5 @@ Wildcard character (`*`) supports including multiple files. For example:
 ```
 
 Files matching the wildcard character are included unsorted. If plugin ordering between files needs to be preserved, the files should be included explicitly.
+
+Environment variables aren't supported in includes section. The file path must be specified as a literal string.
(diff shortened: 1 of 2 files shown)


cosmo0920 (approved):

cosmo0920:
@eschabell Hi, this is not updating to reflect code changes. This describes the current limitation of Fluent Bit's configuration syntax of YAML. 
So, we can :shipit:!

patrick-stephens:
Is this a limitation of just the YAML format? I thought I'd used env vars for includes in classic before. We should call out the delta if so and/or update the classic docs as well.

PR #2302, opened 2025-12-16
out_azure_blob: document path templating

nberlee:
- Documented Azure Blob path templating support added in fluent/fluent-bit#11178.
- Updated the path parameter description plus added a dedicated “Path templating” section with token reference and example.
- Clarified how buffered files retain resolved prefixes

## Summary by CodeRabbit

* **Documentation**
  * Improved Azure Blob output docs: reformatted parameter table for clarity, clarified the path parameter to explicitly support path templating with examples, and added a dedicated "Path templating" section listing supported tokens (strftime directives, millisecond/nanosecond placeholders, UUID, tag references) plus YAML and concrete path examples. No other behavior or defaults changed.


nberlee:
I will put this on draft, there is not a reason for review at this time as the feature is still not merged.

PR #2309, opened 2025-12-19
output: azure_blob: compression with zstd update

nberlee:
Document the new zstd option for Azure Blob network compression and explain how compress_blob inherits its codec, clarifying precedence between transport and blob-level compression.

⚠️ Heads-up: the code enabling Zstandard support is still under review in https://github.com/fluent/fluent-bit/pull/11202, so please keep this docs PR in waiting-on-code-merge until that lands.

## Summary by CodeRabbit

* **Documentation**
  * Expanded Azure Blob output docs to list both gzip and zstd compression options.
  * Clarified compression behavior: defaulting to GZIP when no codec is set and inheriting the chosen codec when compression is configured.
  * Added compatibility note: using compressed blobs with append-blob type yields a configuration error and prevents startup.


pipeline/outputs/azure_blob.md
```diff
@@ -26,8 +26,8 @@ Fluent Bit exposes the following configuration properties.
...
-| `compress`                   | Sets payload compression in network transfer. Supported value: `gzip`                                                                                                                                                        | _none_                        |
-| `compress_blob`              | Enables GZIP compression in the final `blockblob` file. This option isn't compatible when `blob_type` = `appendblob`.                                                                                                        | _none_                        |
+| `compress`                   | Sets payload compression in network transfer. Supported values: `gzip`, `zstd`.                                                                                                                                              | _none_                        |
+| `compress_blob`              | Enables compression in the final `blockblob` file. Defaults to GZIP and inherits the codec from `compress` when that option is set. This option isn't compatible when `blob_type` = `appendblob`.                            | _none_                        |
```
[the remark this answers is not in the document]
nberlee: Updated the compress_blob entry in azure_blob.md:29-33 so the description now explicitly says that gzip is only used when the option is enabled by itself, while the Default column correctly remains _none_.

pipeline/outputs/azure_blob.md
```diff
@@ -26,8 +26,8 @@ Fluent Bit exposes the following configuration properties.
...
-| `compress`                   | Sets payload compression in network transfer. Supported value: `gzip`                                                                                                                                                        | _none_                        |
-| `compress_blob`              | Enables GZIP compression in the final `blockblob` file. This option isn't compatible when `blob_type` = `appendblob`.                                                                                                        | _none_                        |
+| `compress`                   | Sets payload compression in network transfer. Supported values: `gzip`, `zstd`.                                                                                                                                              | _none_                        |
+| `compress_blob`              | Enables compression in the final `blockblob` file. When enabled without `compress`, it uses GZIP; if `compress` is also set, it inherits that codec. This option isn't compatible when `blob_type` = `appendblob`.           | _none_                        |
```
patrick-stephens: What does `isn't compatible` mean here? Does it error out or silently disable or something else?

nberlee: This wasn't changed, fluent-bit will error out as the two are not compatible, so fluent-bot will not be able to know what to do.
see https://github.com/fluent/fluent-bit/blob/master/plugins/out_azure_blob/azure_blob_conf.c#L669

patrick-stephens: Sure, I don't mean anything about change I mean it's not clear what it means so can you update with a bit more detail. If you configure this incorrectly does it trigger an error and fail or carry on but silently ignore.

nberlee: addressed now

PR #2320, opened 2026-01-06
in_tail: add thread ring buffer configuration options

jinyongchoi:
Add `thread.ring_buffer.capacity` and `thread.ring_buffer.window` parameters to the Tail input plugin documentation.

These options are already documented in other input plugins (OpenTelemetry, Prometheus text file). While testing with fluent/fluent-bit#11337, found that these parameters are also useful for tuning the Tail plugin in threaded mode.

When using `threaded: true`, users may need to adjust the ring buffer settings to handle high-volume log ingestion or optimize flush timing based on their workload.

## Documentation
* Added two new configuration options to the Tail input plugin for threaded mode: `thread.ring_buffer.capacity` (ring buffer capacity, default: 1024) and `thread.ring_buffer.window` (flush threshold percentage, default: 5).


alexakreizinger (approved):
thanks for copying these over from the other docs! I didn't realize they applied to this plugin too😁

PR #2366, opened 2026-01-30
lua: fix return value documentation for record and code 2

bplessis-swi:
Lua documentation state that the `record` field is only used when the `code` is equal to `1` but it's also used with `2` when we only change the record content if i am not mistaken

## Summary by CodeRabbit

* **Documentation**
  * Clarified Lua callback docs: when returned codes indicate replacement (now applies to two specific codes) the original record is replaced; clarified timestamp handling and that a callback may emit multiple records if the returned value is an array.


eschabell (approved):
@bplessis-swi this is a good catch, thanks for the PR fix!

PR #2453, opened 2026-03-18
docs: contributing: add plugin documentation standards section

eschabell:
- Document alphabetical parameter table sorting requirement
  - Document workers as a common param to always include
  - Document YAML + classic config tab example structure
  - Document key casing: lowercase for YAML, Title_Case for classic .conf

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,6 +110,53 @@ When you create a new `.md` file for a new page, you must add an entry to this r
 
 Similarly, if you update the `# h1` title header of an existing page, be sure to update that page's `SUMMARY.md` entry to match. `SUMMARY.md` entries takes precedence over in-page headers, which means that if you update a page's `# h1` title without updating `SUMMARY.md`, the unchanged `SUMMARY.md` title will persist in both the rendered page and the table of contents.
 
+## Plugin documentation
+
+When documenting a Fluent Bit plugin, follow these standards to keep plugin pages consistent.
+
+### Configuration parameters table
+
+- **Sort parameters alphabetically** by key name.
+- **List all parameters** defined in the plugin's `config_map`, plus the common `workers` parameter.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -110,6 +110,52 @@ When you create a new `.md` file for a new page, you must add an entry to this r
...
+
+- **Sort parameters alphabetically** by key name.
+- **List all parameters** defined in the plugin's `config_map`, plus the common `workers` parameter.
+- Use this table header format:
+
+  ```markdown
+  | Key | Description | Default |
+  | --- | ----------- | ------- |
+  ```
+
```
patrick-stephens, line 128: Probably add leave the default empty if none is provided, rather than put `none` or similar

CONTRIBUTING.md
```diff
@@ -110,6 +110,52 @@ When you create a new `.md` file for a new page, you must add an entry to this r
...
+...
+{% endtab %}
+{% endtabs %}
+```
+
+### Key casing in examples
+
+Key casing differs between the two configuration formats:
+
+- **YAML** (`fluent-bit.yaml`): All keys are lowercase. For example: `api_key`, `log_level`, `match`.
```
patrick-stephens, line 148: I think we say we match the key name from classic otherwise, using underscores is not valid but we accept it, or we go through and make it all use the camelCase version of the parameter which would be my preference but it's a load of work now. 

There's an open issue on the Fluent Bit side about making the YAML format correct. There's other problems currently as well, e.g. you can define the same key with different cases, with or without underscores, etc. But all out of scope here so I'd say we just keep it consistent with the classic config for now.

patrick-stephens: https://github.com/fluent/fluent-bit/issues/7593 was the one I was thinking of and @stevehipwell may have some follow up, although I don't think it should block the initial contribution guide update. Let's get that in and decide how we want to document YAML after that.

eschabell: Yeah, that's nice, once implemented then we update this doc, can't doc it ahead of time.

CONTRIBUTING.md
eschabell: Fixing.

PR #2550, opened 2026-04-19
docs: outputs: syslog: clarify TLS requires tcp mode

eschabell:
- update mode parameter description to note TLS requires tcp mode and that DTLS over UDP isn't supported

  Applies to code fix in 5.0.3 with no doc PR.


cosmo0920 (approved):

eschabell:
Thanks @cosmo0920 for the review!

PR #2572, opened 2026-05-12
docs: outputs: file: add log rotation support and configuration examples

SagiROosto:
This is the docs PR for https://github.com/fluent/fluent-bit/pull/11352 about `out_file` logrotate feature

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for size-based log rotation in the File output plugin. The config now describes `rotate` with `rotate_max_size` (supports `k/m/g` suffixes), optional gzip compression via `rotate_gzip`, and retention control via `rotate_max_files`. Includes a “Log rotation” section explaining when rotation occurs (on next flush), timestamped/random-hex rotated filenames, oldest-file deletion once limits are reached, and YAML/Fluent Bit CONF examples.

### DIFF
--- a/pipeline/outputs/file.md
+++ b/pipeline/outputs/file.md
@@ -11,7 +11,11 @@ The _File_ output plugin lets you write the data received through the input plug
 | Key | Description | Default |
 | :--- | :--- | :--- |
 | `file` | Set filename to store the records. If not set, the filename will be the `tag` associated with the records. | _none_ |
+| `files_rotation` | Enable size-based [log rotation](#log-rotation). When enabled, files that exceed `max_size` are rotated and optionally compressed. | `false` |
 | `format` | The [format](#format) of the file content. | _none_ |
+| `gzip` | Compress rotated files using gzip. Only applies when `files_rotation` is enabled. | `true` |
+| `max_files` | Maximum number of rotated files to retain per output file. Oldest files are deleted first. Must be `1` or greater. Only applies when `files_rotation` is enabled. | `7` |
+| `max_size` | Maximum size of the active output file before rotation is triggered. Supports size suffixes: `k` (kilobytes), `m` (megabytes), `g` (gigabytes). Only applies when `files_rotation` is enabled. | `100m` |
 | `mkdir` | Recursively create output directory if it doesn't exist. Permissions set to `0755`. | `false` |
 | `path` | Directory path to store files. If not set, Fluent Bit will write the files in its own working directory. | _none_ |
 | `workers` | The number of [workers](../../administration/multithreading.md#outputs) to perform flush operations for this output. | `1` |
@@ -111,6 +115,64 @@ You will get the following output:
 ```text
 1564462620.000254 used=1045448 free=31760160 total=32805608
 ```
+
+## Log rotation
+
+The File output plugin supports size-based log rotation.
+
+When `files_rotation` is enabled, the plugin monitors the size of each output file. Once a file exceeds `max_size`, the next flush rotates the file by renaming it with a timestamp suffix in the format `<filename>.<YYYYMMDD_HHMMSS_XXXXXXXX>`. The `YYYYMMDD_HHMMSS` is machine local timestamp of the time the rotation occurred, and `XXXXXXXX` is a random identifier to guarantee unique filenames if multiple rotations happen within the same second.
+
+If `gzip` is enabled (the default), rotated files are compressed with gzip and stored with an additional `.gz` extension (for example, `cpu.log.20260512_134500_a1b2c3d4.gz`).
+
+The plugin retains up to `max_files` rotated files per output file. When the limit is reached, the oldest rotated files are deleted automatically.
+
+Log rotation works with all supported output [formats](#format): `plain`, `CSV`, `LTSV`, `template`, and `msgpack`. File operations are thread-safe, so rotation can be used alongside multiple [workers](../../administration/multithreading.md#outputs).
+
+### Log rotation example
+
+The following configuration writes CPU metrics to file with rotation enabled. Files are rotated at 50 MB and the five most recent rotated files are retained with gzip compression:
+
+{% tabs %}
+{% tab title="fluent-bit.yaml" %}
+
+```yaml
+pipeline:
+  inputs:
+    - name: cpu
+      tag: cpu
+
+  outputs:
+    - name: file
+      match: '*'
+      path: /var/log/fluent-bit
+      file: cpu.log
+      files_rotation: true
+      max_size: 50m
+      max_files: 5
+      gzip: true
+```
+
+{% endtab %}
+{% tab title="fluent-bit.conf" %}
+
+```text
+[INPUT]
+  Name cpu
+  Tag  cpu
+
+[OUTPUT]
+  Name           file
+  Match          *
+  Path           /var/log/fluent-bit
+  File           cpu.log
+  Files_Rotation true
+  Max_Size       50m
+  Max_Files      5
+  Gzip           true
+```
+
+{% endtab %}
+{% endtabs %}
 
 ## Get started
 
--- a/pipeline/outputs/file.md
+++ b/pipeline/outputs/file.md
@@ -11,13 +11,13 @@ The _File_ output plugin lets you write the data received through the input plug
 | Key | Description | Default |
 | :--- | :--- | :--- |
 | `file` | Set filename to store the records. If not set, the filename will be the `tag` associated with the records. | _none_ |
-| `files_rotation` | Enable size-based [log rotation](#log-rotation). When enabled, files that exceed `max_size` are rotated and optionally compressed. | `false` |
 | `format` | The [format](#format) of the file content. | _none_ |
-| `gzip` | Compress rotated files using gzip. Only applies when `files_rotation` is enabled. | `true` |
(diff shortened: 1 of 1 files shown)


eschabell:
@coderabbitai review

pipeline/outputs/file.md
```diff
@@ -112,6 +116,64 @@ You will get the following output:
...
+
+The File output plugin supports size-based log rotation.
+
+When `files_rotation` is enabled, the plugin monitors the size of each output file. Once a file exceeds `max_size`, the plugin rotates the file by renaming it with a timestamp suffix in the format `<filename>.<YYYYMMDD_HHMMSS_XXXXXXXX>`. The YYYYMMDD_HHMMSS is timestamp of the time when the rotation occurred, and `XXXXXXXX` is a random identifier to guarantee unique filenames if multiple rotations happen within the same second.
+
+If `gzip` is enabled (the default), rotated files are compressed with gzip and stored with an additional `.gz` extension (for example, `cpu.log.20260512_134500_a1b2c3d4.gz`).
+
+The plugin retains up to `max_files` rotated files per output file. When the limit is reached, the oldest rotated files are deleted automatically.
+
+Log rotation works with all supported output [formats](#format), including `out_file`, plain, CSV, LTSV, template, and msgpack. File operations are thread-safe, so rotation can be used alongside multiple [workers](../../administration/multithreading.md#outputs).
```
eschabell: @SagiROosto can you put back-ticks around all the formats listed here: 

- plain
- CSV
- LTSV
- template
- msgpack.

eschabell (changes requested):
@SagiROosto see changes to be made first before review passes.

pipeline/outputs/file.md
```diff
@@ -112,6 +116,64 @@ You will get the following output:
 1564462620.000254 used=1045448 free=31760160 total=32805608
 ```
 
+## Log rotation
+
+The File output plugin supports size-based log rotation.
+
+When `files_rotation` is enabled, the plugin monitors the size of each output file. Once a file exceeds `max_size`, the plugin rotates the file by renaming it with a timestamp suffix in the format `<filename>.<YYYYMMDD_HHMMSS_XXXXXXXX>`. The YYYYMMDD_HHMMSS is timestamp of the time when the rotation occurred, and `XXXXXXXX` is a random identifier to guarantee unique filenames if multiple rotations happen within the same second.
```
eschabell: @SagiROosto also add back-ticks to the "YYYYMMDD_HHMMSS" for consistency.

eschabell (changes requested):

pipeline/outputs/file.md
```diff
@@ -11,7 +11,11 @@ The _File_ output plugin lets you write the data received through the input plug
 | Key | Description | Default |
 | :--- | :--- | :--- |
 | `file` | Set filename to store the records. If not set, the filename will be the `tag` associated with the records. | _none_ |
+| `files_rotation` | Enable size-based [log rotation](#log-rotation). When enabled, files that exceed `max_size` are rotated and optionally compressed. | `false` |
```
patrick-stephens: Just a query on how this is done, is it as soon as they reach that size or a periodic check or something else?

pipeline/outputs/file.md
```diff
@@ -11,7 +11,11 @@ The _File_ output plugin lets you write the data received through the input plug
...
 | `format` | The [format](#format) of the file content. | _none_ |
+| `gzip` | Compress rotated files using gzip. Only applies when `files_rotation` is enabled. | `true` |
+| `max_files` | Maximum number of rotated files to retain per output file. Oldest files are deleted first. Must be `1` or greater. Only applies when `files_rotation` is enabled. | `7` |
+| `max_size` | Maximum size of the active output file before rotation is triggered. Supports size suffixes: `k` (kilobytes), `m` (megabytes), `g` (gigabytes). Only applies when `files_rotation` is enabled. | `100m` |
```
patrick-stephens: What happens if you do not enable rotation? Does it fail to start, report a warning, etc.?

Similarly what if you use invalid values here? Does it fail with a config error, use the default, etc?

pipeline/outputs/file.md
```diff
@@ -112,6 +116,64 @@ You will get the following output:
 1564462620.000254 used=1045448 free=31760160 total=32805608
 ```
 
+## Log rotation
+
+The File output plugin supports size-based log rotation.
+
+When `files_rotation` is enabled, the plugin monitors the size of each output file. Once a file exceeds `max_size`, the plugin rotates the file by renaming it with a timestamp suffix in the format `<filename>.<YYYYMMDD_HHMMSS_XXXXXXXX>`. The `YYYYMMDD_HHMMSS` is timestamp of the time when the rotation occurred, and `XXXXXXXX` is a random identifier to guarantee unique filenames if multiple rotations happen within the same second.
```
patrick-stephens: Where does the timestamp come from? Is it UTC or local timezone based?

Be nice to have a controllable format for the future too.

patrick-stephens (approved):
Some queries around bad-path behaviours but nothing blocking, it would be nice to resolve them though for the inevitable failures that people will have.

pipeline/outputs/file.md
eschabell: @SagiROosto can you answer this for @patrick-stephens ?

pipeline/outputs/file.md
eschabell: @SagiROosto can you answer this for @patrick-stephens ?

pipeline/outputs/file.md
eschabell: @SagiROosto can you answer this for @patrick-stephens ?

eschabell (changes requested):
@SagiROosto can you answer the tagged questions for @patrick-stephens ?

pipeline/outputs/file.md
SagiROosto: TL;DR- size check runs at the start of each output flush for that file, before the new chunk is appended. if that chunk would cause the file to be greater than `max_size` it will rotate it.

The plugin keeps a per-output-file byte counter (`entry->size`), updated from `fstat()` after each successful flush. On flush, if files_rotation is enabled and `max_size > 0`, it compares that counter to `max_size;` when `entry->size >= max_size`, it renames the active file, optionally gzips it, prunes old rotated files, then opens a new active file and writes the current flush there.

On the first flush for a path, the counter is seeded from the on-disk file size (via stat) so a pre-existing file that already exceeds `max_size` is rotated on the next flush.

So rotation is flush-driven, not timer-driven. In practice the active file can exceed `max_size` until the flush that crosses the threshold finishes; the next flush triggers rotation. A single very large flush is not split mid-write—only checked once per flush at the boundary.

pipeline/outputs/file.md
SagiROosto: if `files_rotation` is disabled (default false): Fluent Bit starts normally and the rotation-related configs are ignored.

if `files_rotation` is true but `max_size` is 0 (or "false"): Starts successfully; rotation stays disabled (the code requires max_size > 0). An info log is emitted only when rotation is actually active.

Invalid `max_files` with rotation enabled (`<= 0`): Plugin init fails; Fluent Bit does not start. Will Print `Error: invalid max_files=...; must be >= 1`. Covered by `flb_test_file_logrotate_max_files_validation` in `tests/runtime/out_file_logrotate.c`.

Invalid `max_size` string: Parsed via `FLB_CONFIG_MAP_SIZE` / `flb_utils_size_to_bytes()`. Unlike booleans, an unparseable value is not rejected at config load today—it becomes a very large size_t, so rotation effectively never triggers. Worth documenting as a caveat; a stricter validation could be a follow-up.

Invalid boolean values (`files_rotation`, `gzip`): Standard config-map boolean validation—config error, startup fails.

Runtime rotation failures (rename/gzip): Logged at error/warn; the flush continues and appends to the existing active file (rotation failure does not fail the whole flush). That’s operational behavior rather than config validation.

pipeline/outputs/file.md
SagiROosto: The suffix uses local wall-clock time, not UTC: `time(NULL) → localtime_r() → strftime(..., "%Y%m%d_%H%M%S", ...)`, then `_ `plus 8 lowercase hex digits from `flb_random_bytes()` (fallback if RNG fails) so multiple rotations in the same second don’t collide.

Example: `cpu.log.20260512_134500_a1b2c3d4` (and `.gz` when gzip is enabled).

There is no configurable timestamp format in the current implementation—fixed `YYYYMMDD_HHMMSS` + random tail. Agree a future option (e.g. UTC vs local, strftime template) would be useful; happy to track as a follow-up if we want it called out in the docs “limitations” section.

SagiROosto:
Done, see my replies. @eschabell @patrick-stephens

pipeline/outputs/file.md
patrick-stephens: Thanks, it would be good to capture that somewhere - logrotate and how it works can be sometimes confusing even for me when I've been burnt by it a few times!

patrick-stephens:
Can we capture the clarifications in the actual docs as I know at least some of those questions will likely come up and better than tracking down this PR?

SagiROosto:
@patrick-stephens I've added a `## Log rotation` section that should answer all these questions.
let me know if additional clarifications needed

eschabell (approved):
@SagiROosto thanks for the updates, good to go now once the code PR merges.

patrick-stephens:
DCO check needs to be resolved before we can merge

eschabell:
@SagiROosto code PR merged, merging docs PR.

PR #2611, opened 2026-07-01
filter_kubernetes: Document metadata cache and freshness

Abdessamade03:
## What

Adds a "Metadata cache and freshness" section to the Kubernetes filter documentation, explaining:

- How pod metadata is cached (an in-memory table keyed by `namespace` + `pod_name`).
- Why the default `kube_meta_cache_ttl=0` means entries have no time-based expiry (random eviction only at capacity), so long-lived pods with a stable name (for example, `StatefulSet` members) keep serving the labels and annotations captured when the entry was first cached, until the pod restarts or the entry is evicted.
- How to tune freshness with `kube_meta_cache_ttl`, `kube_meta_namespace_cache_ttl`, and `cache_use_docker_id`, including the request-load tradeoffs.

Also cross-links the "You can't see new objects getting metadata" troubleshooting note to the new section.

## Why

Documentation half of the improvement discussed in fluent/fluent-bit#11961, where the current (by-design) caching behavior surprised a user running StatefulSets. @patrick-stephens suggested documenting the behavior and caveats as an immediately-useful first step; an opt-in implementation improvement will follow separately.

## Notes

- Docs-only change to `pipeline/filters/kubernetes.md`.
- Checked against the repository's `.markdownlint.json` (MD013 disabled, atx headings, `_` emphasis).

## Summary by CodeRabbit

* **Documentation**
  * Added a new section explaining Kubernetes pod metadata caching, including default freshness behavior and how cache capacity impacts eviction.
  * Clarified when updates to pod labels/annotations (especially in-place changes) may appear stale, and how to tune freshness using available cache-related settings.
  * Expanded troubleshooting guidance for partially missing pod metadata, including Pod-specific refresh behavior and a pointer to the new cache details.

### DIFF
--- a/pipeline/filters/kubernetes.md
+++ b/pipeline/filters/kubernetes.md
@@ -384,6 +384,25 @@ pipeline:
 
 The filter can now gather the values of `pod_name` and `namespace`. With that information, it will check in the local cache (internal hash table) if some metadata for that key pair exists. If it exists, it will enrich the record with the metadata value. Otherwise, it connects to the Kubernetes Master/API Server and retrieves that information.
 
+## Metadata cache and freshness
+
+To keep the request rate against the Kubernetes API server (or Kubelet) low, the filter caches the metadata it retrieves for each Pod. On the first record for a given Pod, the filter queries the API server (or the Kubelet, see [Using Kubelet to get metadata](#using-kubelet-to-get-metadata)) and stores the result in an in-memory cache keyed by the Pod's `namespace` and `pod_name`. Subsequent records that resolve to the same key reuse the cached metadata instead of issuing another request.
+
+Because the cache is keyed by `namespace` and `pod_name`, how fresh the labels and annotations are depends on how long an entry stays cached:
+
+- With the default `kube_meta_cache_ttl` of `0`, cache entries have no time-based expiry and are only evicted at random once the cache reaches capacity. For short-lived Pods, such as `Deployment` replicas, this is usually fine: a recreated Pod gets a new name, which is a new cache key, and therefore fresh metadata.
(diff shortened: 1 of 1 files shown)


pipeline/filters/kubernetes.md
```diff
@@ -384,6 +384,25 @@ pipeline:
...
+
+{% hint style="info" %}
+This is expected behavior. The cache trades metadata freshness for a lower request rate against the Kubernetes API server, which matters on large clusters.
+{% endhint %}
+
+If you need labels or annotations that change in place to be reflected sooner, you can tune the cache with the following options, keeping the added request load in mind:
+
+- `kube_meta_cache_ttl`: set a time-to-live, for example `60` or `60s`, so that Pod metadata entries expire and are re-fetched on the next record. Staleness is then bounded by roughly the TTL, at the cost of one extra request per Pod each time an entry expires. Lower values increase freshness and load; higher values reduce both.
+- `kube_meta_namespace_cache_ttl`: the equivalent option for namespace labels and annotations, which defaults to `900` seconds. See [Kubernetes namespace meta](#kubernetes-namespace-meta).
```
patrick-stephens, line 403: I think this resolved the specific statefulset problem right? Might be good to call this out in a specific statefulset use case example showing what happens with and without it

Abdessamade03: Good call, I added a concrete web-0 example showing an in-place kubectl label change: default kube_meta_cache_ttl=0 keeps the old label, kube_meta_cache_ttl 60 picks up the new one on the next record after expiry; also noted cache_use_docker_id for the recreation case.

eschabell (approved):
@Abdessamade03 verified this against the current code base, merging!

PR #2620, opened 2026-07-11
docs: pipeline: inputs: winevtlog: document event_data_as_map option

eschabell:
Document the event_data_as_map and event_template_cache_size options added to the in_winevtlog plugin, which emit EventData as a named map built from provider metadata instead of string inserts.

Note, code commit 6b005e6c4 without corresponding docs PR.

## Summary by CodeRabbit

* **Documentation**
  * Documented the `event_data_as_map` option for Windows Event Log input.
  * Explained how event data is represented as a field-name keyed map when enabled.
  * Added configuration requirements and YAML/config-file examples.
  * Clarified the relationship with template cache sizing and related validation behavior.

### DIFF
--- a/pipeline/inputs/windows-event-log-winevtlog.md
+++ b/pipeline/inputs/windows-event-log-winevtlog.md
@@ -10,31 +10,33 @@ The _Windows Event logs_ (`winevtlog`) input plugin lets you read Windows Event 
 
 The plugin supports the following configuration parameters:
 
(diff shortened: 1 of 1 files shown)


eschabell:
@patrick-stephens ready for review and I've added the exact triggering code commit for you!

cosmo0920 (approved):

PR #2638, opened 2026-07-31
docs: pipeline: outputs: s3: document S3-compatible endpoints

goanpeca:
The `endpoint` description in the configuration parameters table doesn't say what a custom endpoint is for, even though the page later shows MinIO and Google Cloud examples that rely on it. This expands the description to state that `endpoint` points the S3 output at an S3-compatible object store other than Amazon S3, and adds the two behaviors that matter when you do that.

Both behaviors come from the plugin source. `s3_put_object` builds the request URI as `/<bucket><key>` while `cb_s3_init` sets the client host to the configured endpoint, so requests are always path-style rather than virtual-hosted. `cb_s3_init` also sets `s3_client->region` from `region`, which `flb_signv4_do` in `src/aws/flb_aws_util.c` uses for the SigV4 credential scope, so `region` still matters when the endpoint isn't an AWS one.

No other changes: the parameter table stays alphabetical and the existing MinIO and Google Cloud sections are untouched.

## Summary by CodeRabbit

* **Documentation**
  * Expanded S3 endpoint configuration guidance to cover schemes, ports, S3-compatible object stores, path-style requests, and region-based signing.


eschabell:
@goanpeca I've fixed the lint / vale issues, removing draft status as confirmed the fix is good. I'll review now.

eschabell (approved):
@goanpeca this is approved, thanks for the docs PR!

PR #2641, opened 2026-08-01
docs: pipeline: inputs: tail: document windows.path_encoding

eschabell:
Document the Windows-only path encoding mode, including the utf8 value for tailing paths that the active ANSI code page can't represent, extended-length path prefix support, and the distinction from the unicode.encoding content conversion parameter.


eschabell:

pipeline/inputs/tail.md
```diff
@@ -498,6 +499,44 @@ While file rotation is handled, there are risks of potential log loss when using
 
 {% endhint %}
 
+## Windows path encoding
```
patrick-stephens, line 502: I think this has been an oft-requested thing for users so worth covering in upgrade notes.

patrick-stephens (approved):

pipeline/inputs/tail.md
eschabell: Once it releases, we should!

PR #2648, opened 2026-08-01
docs: pipeline: outputs: elasticsearch: document the elasticsearch alias

eschabell:
Note that the es output plugin also loads under the name elasticsearch in Fluent Bit 5.1 and later, and that the alias applies only to output plugins so the elasticsearch input plugin is unaffected.


eschabell:

patrick-stephens (approved):

PR #2650, opened 2026-08-01
docs: administration: yaml: document the extensions section

eschabell:
- Add extensions to the list of supported top-level YAML sections.
- Add an "extensions section" section describing it as a place to keep settings for tooling around Fluent Bit, which Fluent Bit parses and retains but the data pipeline never reads.
- Note that extensions is the only section accepting nested maps, and that nesting elsewhere fails with "variant values are only valid in the extensions section".


eschabell:

patrick-stephens (approved):

PR #2708, opened 2026-09-14
docs: workflows: bump prologic/action-markdownlint from 0.28.0 to 0.30.0

eschabell:
The v0.28.0 action builds its image from a Dockerfile based on
  node:24-bullseye-slim. Debian bullseye is archived, so the apt-get
  install in that build now 404s on perl-modules-5.32, libperl5.32 and
  libnghttp2-14. The build fails before the container starts, so
  markdownlint never runs and the check reports failure regardless of
  the markdown in the PR.

  v0.30.0 uses a prebuilt, digest-pinned image from ghcr.io instead of
  building at job time, so no apt-get runs at all. The github_token and
  reporter inputs are unchanged.

Signed-off-by: Eric D. Schabell <eric@schabell.org>

## Summary by CodeRabbit

* **Chores**
  * Updated the Markdown linting workflow to use a newer validation tool version.


eschabell:
@patrick-stephens this fixes an issue now showing up on https://github.com/fluent/fluent-bit-docs/pull/2707, can you review and merge this?

eschabell:
Closing as the dependabot fix in #2709 solved this.

